PR #43, opened 2015-01-01
Add the Minecraft Wiki information to server-properties.rst

Lemonous:
Continuing my crusade to make things as easy as possible for those who need the docs translated (albeit with no heed to those poor souls who have to do the actual translating, see PR #42), I have added the information about each property found on Minecraft Wiki to this doc page (As a side note, tables are horrible to work with in Sphinx).

If you want to see the styling in practice, try https://livesphinx.herokuapp.com (you might have to zoom out a bit) (Thanks, Andrew).

I also have a few notes to the additions:

1. Is the info about strings on line 59 complete? I don't know if there's a cap on the length, other than the one mentioned in the MOTD-description (lines 208 through 214).

2. Also MOTD-related: Do we want a doc page concerning colours and text formatting codes?

3. There are external links in generator-settings (lines 135 and 137). While this is what I was trying to avoid by including this table, I would also argue that the information in this table is very relevant to those who check this page, whereas the specifics of world generation settings are not. Translations can simply state that more information for those who are interested is available in English on the Minecraft Wiki.

4. Does this make the page too bloated? I thought about making the table its own page, but decided that it was a bit illogical to cut up the information like that.

### DIFF
--- a/source/users/server-properties.rst
+++ b/source/users/server-properties.rst
@@ -8,8 +8,6 @@ Default File
 
 
 Here is the default server.properties file of an unmodified Minecraft 1.8.1 server.
-The function of each property is explained in some detail on the Minecraft Wiki at
-http://minecraft.gamepedia.com/Server.properties
 
 .. code-block:: properties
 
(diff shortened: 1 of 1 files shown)


BlackGoat:
A colour/text-formatting page (2) would be OK, but low priority. As for (3), someone can add a page for it later if there's demand. And (4) it's no more bloated than the MC wiki page (should we credit them somewhere, or did I miss that?). I hope the table looks OK when it gets to RtD... I'll wait to see if anyone else has any comments.

Lemonous:
Yeah, added credit. It's nearly a verbatim copy, after all.

BlackGoat:
OK, as expected, there are problems in the way RtD displays the tables. It's far too wide, and all the text on the end columns is in single lines. Can this be fixed easily, or should I revert this PR? It looks pretty ugly as-is, I'm afraid @Lemonous

BlackGoat:
There is work afoot to alter the RtD theme, which may fix this entirely [these chaps know what I speaketh of: @sk89q @gratimax ]. The only alternative is to use csv formatting, which allows fixed column widths but would require reformatting the entire table. I'll leave it alone pending theme changes.

Lemonous:
I can have a look at it in a few hours. Is the problem the same as on my phone: that the table is far to wide because the text doesn't wrap?

BlackGoat:
Yes. It apparently can't enforce fixed column widths, defaulting to auto-fit with no text wrapping. I think we'd all prefer a global solution rather than a hacky fix, so if the theme is at fault, it needs adjusting.

Lemonous:
I would think that's the case, since it doesn't break like the Heroku app did (i.e. viewport not wide enough, pushing the last column to the next line and screwing it up completely). The text is also wrapped in the code.

Lemonous:
Did you say you tried something like ``.. tabularcolumns:: |l|l|l|l{300px}|``? I believe that only applies to LaTeX documents. Try using ``.. htmlonly::`` instead, perhaps?

PR #45, opened 2015-01-03
Added package advice.

ryantheleach:


BlackGoat:
Good call, very helpful.

PR #46, opened 2015-01-03
Change "tile" to "block"

Lemonous:
Since "tile entities" were changed to "block entities" with the 1.8 update, and Sponge will be made for 1.8, I find it more natural to change the entries here to "block entities".

Source: http://minecraft.gamepedia.com/Chunk_format#Block_entity_format

### DIFF
--- a/source/plugins/blocks.rst
+++ b/source/plugins/blocks.rst
@@ -42,24 +42,24 @@ In the future, there will only be one 16 bit number (65536 possible combinations
     5 => minecraft:planks[variant=spruce]
     etc.
 
-Tile Entity Data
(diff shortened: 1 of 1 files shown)


Kobata:
I'm honestly not aware of anyone who actually calls them that except for the wiki (and basically just that page) -- the save format still calls it 'TileEntity', etc.

bensku:
Well, Minecraft itself has some debug messages using BlockEntity as name. Forge still uses TileEntity, so I think changing that would be bad idea.

BlackGoat:
Unless there's reformatting in the save format and Forge, this PR isn't going to help.

PR #49, opened 2015-01-11
Eclipse plugin guide

boformer:
Info about gradle and git plugins for eclipse.

### DIFF
--- a/source/plugins/eclipse.rst
+++ b/source/plugins/eclipse.rst
@@ -3,13 +3,44 @@ Eclipse
 =======
 
 Eclipse for Java
-~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~
 Eclipse is a development environment that you can learn and code Java with. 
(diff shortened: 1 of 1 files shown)


Mumfrey:
To be honest you should point out that if people install the correct version of Eclipse in the first place (Eclipse for Java Developers) then egit is built in.

boformer:
I did not know that ;)

source/plugins/eclipse.rst
```diff
@@ -3,13 +3,44 @@ Eclipse
...
+
+Eclipse supports many plugins that simplify development tasks, especially for beginners.
+
+To install plugins, go to ``Help >> Install New Software...`` and enter the the update site location.
+Alternatively, you can drag the 'Install Button' from the Eclipse marketplace into your running Eclipse instance.
+
+Eclipse Gradle Integration
+--------------------------
+
+This plugin adds a visual interface for Gradle, the dependeny management and build system used by the Sponge Project and many plugin developers.
```
BlackGoat: misspelled 'dependency'

PR #51, opened 2015-01-11
how to access and provide services

boformer:
Content is based on the javadoc and this status update:
https://forums.spongepowered.org/t/status-update-26-september-2014-api-developer-form/2859

### DIFF
--- a/source/plugins/services.rst
+++ b/source/plugins/services.rst
@@ -1,3 +1,29 @@
 ==================
 Accessing Services
 ==================
+Pretty much everything (events, permissions, etc.) is handled through services. 
+All services are accessed through the service manager:
+
+.. code-block:: java
+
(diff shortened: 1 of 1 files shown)


boformer:
maybe it would be better to use the EventManager instead of PermissionService in the first example?

The events page is linked to the services page:
https://docs.spongepowered.org/en/latest/plugins/event-bus/

ryantheleach:
It should be whatever is expected to be more common or simple.

It should provide a clear example of how to get the common sponge provided services, but also contain an example of how to provide your own service based on your own or a third party interface (think vault for economy)

boformer:
I added a simple example and replaced EventManager with PermissionService

BlackGoat:
I haven't had any negative feedback from @sibomots so I'm going to merge this.

PR #54, opened 2015-01-12
Make Migrating from Bukkit more clear

lexware:


boformer:
I don't know how far the Pore project is, but a reference would be great.

boformer:
https://github.com/LapisBlue/Pore

lexware:
@boformer I didn't mention Pore because there are no publicly available builds, I didn't want to create confusion.

BlackGoat:
The text looks rather cumbersome in English. How about
"No, unfortunately there is no native support for Bukkit plugins in Sponge. (It may be possible to re-implement the Bukkit API with a special Sponge plugin, and thus support Bukkit plugins, but this has not yet been demonstrated.)"

lexware:
Ok, I will start to reword it, as a developer it made perfect sense.

—

PR #55, opened 2015-01-17
Add 'Migrating from Canary' page

lexware:


boformer:
What if we make only one "Migration" page for Bukkit, Canary, Spigot, Cauldron etc.?

lexware:
The information isn't the same for all eg Canary uses the normal world storing.

boformer:
That's right, there would be different sections for different server mods. 

I just think that it is not a lot of information, so we can combine the pages.

lexware:
Oh that's a good idea, Bukkit and Spigit are pretty much the same by the way and conversion from cauldron is just removing cauldron.

—

BlackGoat:
So the topic is more like "Migrating TO Sponge" now? It's a good idea, and the sections can be easily found using a TOC at the top.

BlackGoat:
Looking at this harder, the Compat page branches to Bukkit and Forge pages. Adding one more won't hurt for now. It can all get condensed into Compat later if needed. Cauldron can probably get mentioned on the Forge page.

PR #61, opened 2015-01-20
plugins.rst - Mentioned implementation specific plugins.

ryantheleach:
Plugins being implementation specific was the most common source of problems between updates for Bukkit plugins, it deserves to be mentioned in anticipation of it happening on sponge as well.

### DIFF
--- a/source/users/plugins.rst
+++ b/source/users/plugins.rst
@@ -26,6 +26,7 @@ Common Problems
 
 If you are having problems with a plugin, consider the following things:
 
-* If the plugin compatible with your Minecraft version? While Sponge tries to keep old plugins working, this is sometimes not possible.
+* Is the plugin compatible with your Minecraft version? While Sponge tries to keep old plugins working, this is sometimes not possible, (however in most cases plugins based on a stable release should continue functioning without needing to keep up to date).
 * Do you have a new enough Java version? While Sponge is built for Java 6 minimum, some plugins may require newer versions of Java such as 7 or 8.
 * The plugin may be outdated. Is there a newer version of the plugin?
+* Some plugins may be "implementation specific" they may hook into minecraft/forge/glowstone bypassing the SpongeAPI, or otherwise rely on details that change between versions or platforms.


BlackGoat:
This opens a massive can of worms. I made issue #62 to help resolve it. I appreciate your intentions, but I'm not sure I should merge this until that's been discussed more...

boformer:
I think it will confuse people.

We can add this when there are many of these plugin-mod-mixes and it becomes an issue, but right now it is not necessary.

ryantheleach:
I disagree. a lot of people won't re-read the documentation if they read it at all. Either it should be in (probably worded better then I had it) or it shouldn't be mentioned.

progwml6:
as far as i'm concerned in 99% of cases implementation specific plugins are nothing but API deficiencies

gratimax:
@progwml6 then we can mention here to turn them into feature requests

modwizcode:
I would draw attention to this https://github.com/SpongePowered/SpongeAPI/issues/364 which deals with something similar.

BlackGoat:
As this page is intended for Users, any suggestions of API feature requests are out of place.
The real issue here is Implementations - that is, if a plugin hooks into Forge, or Granite, or XYZ. If they do, users need to know - preferably when they DL them. Stuff from Ore might also use Forge, things that hook into Granite will break on Forge, and vice versa. Reliance on a specific build range of Forge/XYZ is possible too. This won't arise for a while, and hopefully when it does we can strong-arm the required functionality into the API, and add something like "Does the plugin need a specific Implementation and/or Build of Sponge? Check at the site you DLed it from. We can't help."

source/users/plugins.rst
```diff
@@ -26,6 +26,7 @@ Common Problems
 
 If you are having problems with a plugin, consider the following things:
 
-* If the plugin compatible with your Minecraft version? While Sponge tries to keep old plugins working, this is sometimes not possible.
+* Is the plugin compatible with your Minecraft version? While Sponge tries to keep old plugins working, this is sometimes not possible, (however in most cases plugins based on a stable release should continue functioning without needing to keep up to date).
 * Do you have a new enough Java version? While Sponge is built for Java 6 minimum, some plugins may require newer versions of Java such as 7 or 8.
 * The plugin may be outdated. Is there a newer version of the plugin?
+* Some plugins may be "implementation specific" they may hook into minecraft/forge/glowstone bypassing the SpongeAPI, or otherwise rely on details that change between versions or platforms.
```
hmksq: `minecraft`, `forge` and `glowstone` isn't capitalised.

Issue #31, opened 2014-12-23
Working with the Message API


BlackGoat:
I have made a stub file for explaining the Message API at source/plugins/message. Please help elucidate this file if you are able.

Issue #20, opened 2014-12-20
HOCON page needs more info


BlackGoat:
The HOCON info page at source/users/hocon needs a bit more friendly text, and an example or two. Folk will want to know how does HOCON format differ from YAML or XML?

PR #64, opened 2015-01-25
Update message.rst, hocon.rst, and event-bus.rst

tyrannokapi:
* Updated source/plugins/message.rst as per #31 
* Updated source/users/hocon.rst as per #20 
* Fixed misspellings of "recipient" in source/plugins/event-bus.rst (formerly "receipient")

### DIFF
--- a/source/plugins/event-bus.rst
+++ b/source/plugins/event-bus.rst
@@ -132,7 +132,7 @@ If you want your event to be cancellable, the class must also implement ``Cancel
    
        public PrivateMessageEvent(Player sender, Player recipient, String message) {
           this.sender = sender;
-          this.recipient = receipient;
+          this.recipient = recipient;
           this.message = message;
        }
     }
--- a/source/plugins/event-bus.rst
+++ b/source/plugins/event-bus.rst
@@ -104,7 +104,7 @@ If you want your event to be cancellable, the class must also implement ``Cancel
        private boolean cancelled = false;
       
        private Player sender;
-       private Player receipient;
+       private Player recipient;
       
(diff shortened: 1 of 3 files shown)


ryantheleach:
Looks good, only thing that I would be concerned about is if sponge is actually going to use HOCON or not. there were talks of changing the ConfigAPI, but it may still end up using HOCON depending on the decision.

tyrannokapi:
@ryantheleach Hmm, I wasn't aware of that. I suppose we'll just have to wait and see.

@boformer I believe the color is appended following the text, came across that when I was cross-referencing the forums. I could be wrong, though.

Thanks!

ryantheleach:
@boformer https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/text/message/Messages.java#L56

Messages.builder("string") creates a new text builder, it's fine.

@tyrannokapi the color doesn't matter whether it's applied before or after the text as it is a single message segment. append creates a new text element and makes it a parent of the root message element.

The only thing really missing that I can see is translatable messages.

boformer:
@ryantheleach I know what Messages.builder("string") does. the question is what messageBuilder.color(...) does.

ryantheleach:
It adds a color component to the builder part representing that piece of text. It doesn't need to be at the front or the end it just works, in order to get multiple colors you need to nest multiple messages using append, unless they are arguments for a translatable message, in which case each argument is treated as a separate format-table message segment.

It's designed to be compatible with vanilla minecrafts json chat format described here: http://wiki.vg/Chat

PR #66, opened 2015-01-26
Add content to the Command Registration page

Maenni025:
Added an example command to the Command Registration page and added some explanation.

### DIFF
--- a/source/plugins/commands.rst
+++ b/source/plugins/commands.rst
@@ -2,3 +2,76 @@
 Command Registration
 ====================
 
+Create a command
+================
+
+The first step is to create a class for the command. The class has to implement ``CommandCallable``:
+
+.. code-block:: java
+
+    package com.github.maenni025.spongeExample;
+
+    import java.util.Collections;
+    import java.util.List;
+
+    import org.spongepowered.api.Server;
+    import org.spongepowered.api.entity.player.Player;
+    import org.spongepowered.api.util.command.CommandCallable;
+    import org.spongepowered.api.util.command.CommandException;
+    import org.spongepowered.api.util.command.CommandSource;
+
+    import com.google.common.base.Optional;
+
+    public class ExampleCommand implements CommandCallable {
+
+        private final Server server;
+
+        public ExampleCommand(Server server) {
+            this.server = server;
+        }
+
+        public boolean call(CommandSource source, String arguments, List<String> parents) throws CommandException {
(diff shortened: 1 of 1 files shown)


source/plugins/commands.rst
```diff
@@ -2,3 +2,76 @@
...
+Now we can register the class in the CommandService
+
+.. code-block:: java
+
+    CommandService cmdService = game.getCommandDispatcher();
+    cmdService.register(plugin, new ExampleCommand(server), "message", "broadcast");
+
+The command is now registered.
+The last argument are the aliases for the command. You can add as many Strings as you want.
+If a alias is used by another command it is ignored. The first alias, which isn't used by another command, becomes the primary alias.
```
12AwsomeMan34: an alias*

source/plugins/commands.rst
```diff
@@ -2,3 +2,76 @@
...
+    public class ExampleCommand implements CommandCallable {
+
+        private final Server server;
+
+        public ExampleCommand(Server server) {
+            this.server = server;
+        }
+
+        public boolean call(CommandSource source, String arguments, List<String> parents) throws CommandException {
+            for (Player p : server.getOnlinePlayers()) {
```
boformer: use server.broadcastMessage(Message message) instead?

source/plugins/commands.rst
```diff
@@ -2,3 +2,76 @@
...
+
+    import org.spongepowered.api.Server;
+    import org.spongepowered.api.entity.player.Player;
+    import org.spongepowered.api.util.command.CommandCallable;
+    import org.spongepowered.api.util.command.CommandException;
+    import org.spongepowered.api.util.command.CommandSource;
+
+    import com.google.common.base.Optional;
+
+    public class ExampleCommand implements CommandCallable {
```
boformer: Rename to MyBroadcastCommand?

source/plugins/commands.rst
```diff
@@ -2,3 +2,76 @@
...
+            }
+            return true;
+        }
+
+        public boolean testPermission(CommandSource source) {
+            return source.hasPermission("example.exampleCommand");
+        }
+
+        public Optional<String> getShortDescription() {
+            Optional<String> desc = Optional.of("Displays a message to all players");
```
boformer: You could return these directly, or store them in a field, so you don't have to initialize a new Optional every time.

Issue #25, opened 2014-12-22
Plugin Config page needs writing


BlackGoat:
The page at source/plugins/configuration is intended for discussion of working with Configuration and defaults in developing Plugins. From the original Wiki plan. Someone please write this!

PR #74, opened 2015-01-28
Update configuration.rst: Added content

tyrannokapi:
I've added sections on getting, setting, and retrieving default configuration values.

I can change as necessary, specifically:
* Need feedback on the section concerning editing/setting default configuration values.
* Usage of withFallback() when editing values - is that something that should be included in this?

Edit: Tackles Issue #25

### DIFF
--- a/source/plugins/configuration.rst
+++ b/source/plugins/configuration.rst
@@ -2,4 +2,144 @@
 Configuration and Defaults
 ==========================
 
-How to work with and read configuration and defaults.
+Plugins using the Sponge API have the option to use one or more configuration files. Configuration files allow plugins to store data, and they allow server administrators to customize plugin options (if applicable).
+
+
+.. _getting-default-config:
+
+Getting your Default Plugin Configuration
+=========================================
+
+The Sponge API offers the use of the ``@DefaultConfig`` annotation on a field or method with the type ``File`` to get the default configuration file for your plugin.
+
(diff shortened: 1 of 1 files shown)


source/plugins/configuration.rst
```diff
@@ -2,4 +2,141 @@
...
+
+     if (!getDefaultConfig().exists()) {
+        File newConfig;
+        try {
+            getDefaultConfig().createNewFile();
+        } catch (IOException exception) {
+            getLogger().log(Level.SEVERE, "The default configuration could not be created!");
+        }
+        ConfigFile config = ConfigFile.parseFile(getDefaultConfig());
+        config.withValue("plugin.version", ConfigValueFactory.fromAnyRef(1));
```
gratimax: This doesn't work because config methods are immutable. Consider them to be like a builder on the same object.

Instead, something like
```java
ConfigFile config = ConfigFile.parseFile(getDefaultConfig())
    .withValue("plugin.version", ConfigValueFactory.fromAnyRef(1));
    .withValue("plugin.doStuff", ConfigValueFactory.fromAnyRef(true));
    .withValue("plugin.doMoreStuff", ConfigValueFactory.fromAnyRef(false));

config.save(true);
```

gratimax:
The misuse of the configuration API confirms some suspicions on how it's unnatural to use because it's immutable. Pinging @sk89q to see if this will change

tyrannokapi:
Whoops, my bad. Thanks for catching that. 

I actually need to change the logger in this PR as well, as I'm using the wrong logger.

Issue #15, opened 2014-12-19
Getting a Logger: description needed


BlackGoat:
On the page source/plugins/logging, the section "Getting a Logger" and below is blank. Could some well-informed developer please complete this section?

PR #75, opened 2015-01-28
Update logging.rst: Added content

tyrannokapi:
Added content on getting a logger and emitting messages as per #15. Some parts are based on comments from another (closed) PR. I can change this as necessary.

### DIFF
--- a/source/plugins/logging.rst
+++ b/source/plugins/logging.rst
@@ -2,14 +2,95 @@
 Logging and Debug Messages
 ==========================
 
-Java comes with a few logging frameworks. Logging is preferrable to printing to stdout (or stderr) with ``System.out.println()`` for a number of reasons:
+There are a few logging frameworks available for use in Java. Logging is preferable to printing to stdout or stderr with ``System.out.println()`` for a number of reasons:
 
-* Logged messages have a source name attached to them so it is possible to figure out where the logged messages are coming from.
-* Logged messages have a severity level which allows for simple filtering (i.e. disable all non-critical notices).
+* Logged messages are labeled with a source name, making it easier to figure out where the logged messages are coming from.
+* Logged messages have a severity level which allows for simple filtering (e.g. disable all non-critical notices).
 * The available logger frameworks allow you to enable or disable messages from certain sources.
+
+Sponge uses ``org.slf4j.Logger``, not ``java.util.logging.Logger``.
 
 Getting a Logger
 ================
 
+The Guice module used during the initialization of plugins has a plugin-scoped logger. This allows you to annotate a field, method, or constructor with ``@Inject`` to get the logger for your plugin, which is pre-configured with the correct plugin ID.
+
+.. note::
+    See :doc:`quick-start` for information on configuring your plugin ID.
+
+**Example - Field**
+
+.. code-block:: java
+
+    import com.google.inject.Inject;
+    import org.slf4j.Logger;
+
+    @Inject
+    private Logger logger;
+
+**Example - Method**
+
+.. code-block:: java
+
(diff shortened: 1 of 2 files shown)


source/plugins/logging.rst
```diff
@@ -2,14 +2,73 @@
...
+    import java.util.logging.Level;
+    import java.util.logging.Logger;
+
+    getLogger().log(Level level, String message);
+
+The ``level`` is the severity of the log message, and can be any field specified within the ``java.util.logging.Level`` class. The ``message`` is the message you want to be shown. For example:
+
+.. code-block:: java
+
+    getLogger.log(Level.WARNING, "Initialization of this one thing failed!");
```
simon816: Missing parameter brackets `()` on getLogger.

Personally I'd store the logger in a `public static final` field so it can be accessed from any class (but that doesn't really matter for the example)

simon816:
Should there be a mention of using `LoggerFactory.getLogger("ExamplePlugin")` as well?

source/plugins/logging.rst
tyrannokapi: Whoops. Thanks, not sure how I missed that. 

Normally, I would too, though I feel that's something that would be better discussed in another guide because it's applicable to more than loggers.

progwml6: We are NOT using java.util.logger in sponge!

tyrannokapi: Corrected, thanks.

source/plugins/logging.rst
```diff
@@ -2,14 +2,75 @@
...
+
+    The following two examples assume that the getter method for your logger is named ``getLogger()``, as shown in the previous section. This may differ for you depending on what you named your getter method.
+
+.. code-block:: java
+
+    import org.slf4j.Logger;
+    import org.slf4j.Marker;
+
+    getLogger().debug(Marker, String);
+    getLogger().error(Marker, String);
```
ryantheleach: I would sort these in order of priority.

progwml6:
lets leave this stuff simple. there is no need to have markers listed in the 1st example

tyrannokapi:
@simon816 Perhaps, but I think that should probably be left out for simplicity.
@progwml6 I agree, I've removed the markers.

source/plugins/logging.rst
```diff
@@ -2,14 +2,77 @@
...
+**Example - Field**
+
+.. code-block:: java
+
+    import com.google.inject.Inject;
+    import org.slf4j.Logger;
+
+    @Inject
+    private Logger logger;
+
```
gratimax: it is also possible to inject it in the constructor, that might be worth mentioning

gratimax:
@tyrannokapi Only problem I see is what I just commented. Eagerly waiting a fix and then merge :)

tyrannokapi:
@gratimax Will do once I get home!

gratimax:
@tyrannokapi great! Is there a change you could merge in e2f5283da6afe6 as well, just to test PR building? As you see, [currently the PR is failing](https://travis-ci.org/SpongePowered/SpongeDocs/builds/48821784) even though there are no errors

ryantheleach:
In a plugin, if you need to access a logger from multiple classes should you a. Keep a reference to your plugin and and do plugin.log(string) which which then calls its logger? B. Get the logger for each class that uses it. C. Use guava to construct all classes that use loggers loggers and inject it.

tyrannokapi:
@gratimax Small typos are fun, heh. Gonna have to fix the rebasing error tomorrow, extra commits got pushed into my branch by accident.

@ryantheleach Personally, I would do Option A out of those three.

PR #88, opened 2015-02-05
Fix table wrap. Resolves #87

### LINKED ISSUE #87: Table width unconstrained on RtD
Tables look terrible on the Read-the-Docs pages (ie. http://sponge.readthedocs.org/en/latest/users/server-properties/ ). Something is wrong with the wrapping on the last column, that does not happen on the GitHub view. Can we fix this with Sphinx or RtD style settings?

hawtre:
Re-adds html static path, adds new theme override stylesheet

### DIFF
--- a/source/_static/rtd_overrides.css
+++ b/source/_static/rtd_overrides.css
@@ -1,0 +1,3 @@
+html body div.wy-table-responsive table td, html body div.wy-table-responsive table th {
+    white-space: normal;
+}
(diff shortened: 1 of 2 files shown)


hawtre:
Do we want to override the css like this?

boformer:
A standalone css file for one small change is very inefficient...

Otherwise, I think this is ready ;)

BlackGoat:
I know @gratimax has a new Style in the works that resolves this bug, but it isn't complete yet. When that rolls out this PR might be reverted, but that may be a while yet (he has other priorities).

PR #90, opened 2015-02-07
Fill in recently added Sponge commands

12AwsomeMan34:
Adds some of the commands.
Should the parameters be specified for commands? The rest of the commands do not include them so I did not. But commands such as `/sponge conf` would be much more useful if the reader knew the parameters.


hawtre:
I think we would want the full syntax for all the commands, detailing their flags and accepted parameters

tyrannokapi:
Looks good.

Parameters would be nice, though the table formatting in RtD may become an issue because of the length.

BlackGoat:
Parameters are needed, it would be very odd of us to omit them from the Docs. If the table ends up looking awful in our present style, it will have to revert to a list or something.

PR #91, opened 2015-02-07
Add a special notice about world configs

12AwsomeMan34:
The configuration tree for worlds was missing (at least I think it is). This knowledge may prove useful.

### DIFF
--- a/source/users/config.rst
+++ b/source/users/config.rst
@@ -37,7 +37,7 @@ There are three types of world configs:
 * World
 
 Global configuration files can affect all of a server's worlds and dimensions. This is the default level for configs.
-Dimension configuration files are used by mods to affect a certain mod world or group of worlds. These types of configs will override the global config files.
+Dimension configuration files are used to affect a certain dimension or group of worlds. These types of configs will override the global config files.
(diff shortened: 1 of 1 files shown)


source/users/config.rst
```diff
@@ -27,5 +27,18 @@ What You Can Configure
...
+======================
+
+There are three types of world configs:
+
+* Global
+* Dimension
+* World
+
+Global configuration files can affect all of a server's worlds and dimensions. This is the default level for configs.
+Dimension configuration files are used by mods to affect a certain mod world or group of worlds. These types of configs will override the global config files.
```
hawtre: Are they only used by mods?

boformer:
You should mention that it is possible to change config values ingame:
```
/sponge conf -d nether logging.chunk-load true
/sponge conf -w DIM1 logging.chunk-load true
```

BlackGoat:
If you add the /sponge conf parameters and examples from https://github.com/SpongePowered/Sponge/commit/be08be04b06b413686c7f6c574bf89e918e1d6d3 to this page, with a reference to the commands page, that would save us from reproducing it there.

PR #93, opened 2015-02-08
Set maven dependency scope to provided.

mbax:
It is expected that the Sponge API would be available at runtime, without having to include it in our plugin jar.

Sorry, I don't have the way to do this in Gradle. :)


boformer:
I think that's the default behaviour of Gradle

mbax:
Great, that makes life easier.

PR #105, opened 2015-02-24
Remove unnecessary upper case letters

jamierocks:


drtshock:
Headers are supposed to be capitalized.

boformer:
@jamierocks http://english.stackexchange.com/questions/14/which-words-in-a-title-should-be-capitalized

PR #106, opened 2015-02-24
Change wording in canary.rst

jamierocks:


modwizcode:
This is too small a change, stop by on IRC and have a docs person roll it into a commit. This isn't large enough, or even correct enough to add to the docs.

PR #110, opened 2015-02-25
Edit description for 64-bit Java.

Xemiru:
- Remove recommendation to use 64-bit Java.
  > Recommendation isn't really something needed there. System architecture is what'll probably dictate the user the most about what to get, and adding a recommendation there may only mislead users to thinking they can choose either one. And that's a problem, because you can't use 64-bit Java on a x32/x86 computer.


progwml6:
users regularly grab the wrong java for their architectures.

Xemiru:
then its probably better to specifically tell them to check based on that instead of recommending one and implying there's a choice where there might not be

>also, i blame java for typically defaulting to giving 32bit java when using a 64bit com .c.
unless that's firefox's fault or something

PR #113, opened 2015-02-27
Remove capitalization from "Server"

Brycey92:
A server is not a proper noun.


drtshock:
+1

gratimax:
Apparently it's too hard to look at the commit history, so  0266312

PR #117, opened 2015-02-27
Improve command example, add comments. Fixes #109

### LINKED ISSUE #109: Improve command class example
The command class example (``plugins/commands.rst``) should be improved.

According to the ``CommandCallable`` javadoc, the usage String should only be ``"<message>"``.

The permission string should be something like ``"example.broadcast"``.

Also, some comments explaining the fields and methods (especially ``call``, ``testPermission`` and ``getSuggestions``) would be great.

boformer:
Fixes #109


gratimax:
Superseded by #122

PR #146, opened 2015-03-23
Quickfixes

tyrannokapi:
This fixes a couple minor issues that have been mentioned:

* The text under "Preparing for Development" was poorly worded for translatability
* The bullets on the Implementations page were improperly formatted

I would like for this PR to stay open for a couple days in case of any other minor errors, so I suppose it could use a WIP label.

### DIFF
--- a/source/about/faq.rst
+++ b/source/about/faq.rst
@@ -5,32 +5,32 @@ Frequently Asked Questions
 For Everyone
 ============
 
-Are there downloads available yet?
+Are There Downloads Available Yet?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 No. When we are ready for a release, an announcement will be posted on the Sponge forums. Keep in mind that Sponge will first be available for Minecraft 1.8.
 
-Why is the official Sponge implementation built on Minecraft Forge?
+Why Is the Official Sponge Implementation Built on Minecraft Forge?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Reusing existing efforts in the community allows us to have a working version much quicker. It would be very time-consuming to write a new server from scratch, and many people are not in a position to wait.
 
-What makes Sponge different from other projects, such as Bukkit?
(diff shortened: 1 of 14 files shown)


Tzky:
On the index page, section "preparing for development":
"Contributing to Sponge." should be  "Contributing to Sponge". 

Its just a small change, so i assume it belongs here. Changing “Creating Plugins” and “Contributing to Sponge" to wirking links might be a solution too.

tyrannokapi:
Yes, you are correct about the period. I think that linking to the appropriate section is a good solution.

The comma before the conjunction in a list is grammatically correct, at least in the US. I believe it is referred to as an Oxford comma. The US may be in the minority with this usage, though... @BlackGoat, do you have a preference on this?

Dannyps:
[The Wikipedia article about the comma](https://en.wikipedia.org/wiki/Serial_comma).

btw, in [this page](https://github.com/SpongePowered/SpongeDocs/blob/master/source/server/getting-started/installation.rst), there's another period out of place. A quick search in Crowdin revealed that these are the only two case in which this happens, so, there shouldn't be any other string with this problem.

BlackGoat:
Being Australian, I tend to favour UK grammar and spelling. Also from Wiki: "The Oxford comma tends not to be used in non-academic publications unless its absence produces ambiguity.". I prefer this fuzzy line - leave it out unless it looks weird (or is easily misinterpreted).

BlackGoat:
Dannyps also spotted this: In the new FAQ, titles are not in Title Case

Dannyps:
in PR #151 there are two sentences that are a bit odd IMO.


**"A full list of staff members may be located at :doc:`staff`."**
could be perhaps:
 - "A full list of staff members **is located** at :doc:`staff`."
 - "A full list of staff members **can be found** at :doc:`staff`."

and

**"This means that server owners run Sponge plugins on any implementation of the Sponge API..."**
laks something... It could be:
 - "This means that server owners **can** run Sponge plugins on any implementation of the Sponge API..."

both strings are from the file 'source/about/introduction.rst'.

I don't know if this PR can be used to correct these problems, but I'll leave them here anyway.

tyrannokapi:
I'll correct those. Thanks.

tyrannokapi:
I believe this is ready.

PR #160, opened 2015-04-10
Update future.rst

TheDoctorsLife:
Make the "if" bold, as I feel that point should be emphasized.


thomas15v:
+1 :thumbsup:

progwml6:
not needed

PR #162, opened 2015-04-21
Trying to fix #156

### LINKED ISSUE #156: Granite merger cleanup
Adding Granite to the mix means some pages must change. Here are the ones I could identify, there may be more I missed. Minor changes in wording are needed in most cases. 

    Index - add fourth component (Granite) - **DONE**
    about/FAQ - There were Downloads (of Granite), etc - **DONE**
    Add /contributing/granite folder and files
    server/getting-started/migrating/forge
    server/getting-started/launch-script
    server/getting-started/installation
    server/management/plugins (Granite vs Forge)
    downloads *
    changelog  *
* It is possible that some of these might need to become two pages, depending on what happens to Granite forums and where its' downloads end up.

Tzky:
First try to use Github properly.
Might get removed, just testing. :)


Tzky:
Nope, that's not how this works.

PR #170, opened 2015-04-26
Fix up downloads in FAQ.

ryantheleach:
A few bits of information were outdated due to the new dev builds available.


Tzky:
Looks good.

PR #176, opened 2015-04-28
Ew GitHub for Windows <3 SourceTree

jamierocks:


Tzky:
What about providing both on the docs? 
Decisions on software choices should be made by users not authors...

gabizou:
Or, you know, could always just direct people to the [Git] website for downloads so they can have infinitely more options than we can advise.

[Git]:www.git-scm.com

Zidane:
What @gabizou said.

jamierocks:
I totally agree with directing people to the Git website, however I assumed a decision against that was made. :/

ryantheleach:
Last time I used github for windows I struggled to do anything remotely useful or complex with it.

TortoiseGIT / GITgui / GIT / Sourcetree all at least give you more control, but my personal favourite is also SourceTree.

I don't really think we need git tutorials in the API, but at the same time I can imagine common scenario's that come up when developing with/for sponge that we should cover...

gabizou:
I would just say "the Git website has a lot of documentation and provides plenty of options for GUI clients." We don't want to be providing git tutorials when there are so many out there, let alone that there is a substantial amount of documentation with using Git.

BlackGoat:
Link to Git website added with explanation. Closing this.

PR #182, opened 2015-05-02
Added Client Side Modding to FAQ.

ryantheleach:


BlackGoat:
Travis and his fixation with underscore length.... but that is a rather long title.

PR #188, opened 2015-05-03
Add figures / shared mixin pages

Tzky:
This PR adds add figures to faq.rst and /contributing/index.
It creates shared mixins pages in contributing/vanilla, ../common and ../sponge too.


gratimax:
While nice, the "sponge-layout" image doesn't have any definition in terms of x or y. The first image does: y is closeness to plugins / Minecraft impl and x is differing implementations. There's no meaning to the axes in the second image, which means that either it doesn't belong as an image or belongs as a different visualization. I can only see it as confusing people.

boformer:
The second image can be replaced with a simple bullet list. I also don't think that "GitHub" should be the parent node of the projects.

Tzky:
I agree with both of you. Removed the 2nd image and using a bullet list for now. I chose Github to make translations as easy as possible in the first place, as images can't be translated properly.

PR #195, opened 2015-05-05
Improve Text API examples

RobertHerhold:
This pull request is a continuation of #191. The changes are:
* Rename all instances of "message" to "text" (where applicable, not a find and replace)
* Add examples for styled text (previously there was only colored text)
* Add an example for shortcut text construction using `Texts#of(Object... objects)`
* Rename messages.rst to text.rst (I can revert this if necessary)
* Remove Text creation from methods


Minecrell:
Don't place text creation in a method, if it's the same every time you can simply store it in a constant

RobertHerhold:
The previous examples had them in methods, so I was just continuing the trend

Minecrell:
I never said the previous docs were perfect :P It's better to fix this now, creating methods with identical behavior every time is pretty much pointless, and Text is designed to be able to be stored in constants efficiently.

RobertHerhold:
Sounds good :) I'll update them.

PR #207, opened 2015-05-09
Explain naming of Sponge coremod artifacts. 

### LINKED ISSUE #194: Mis-matching Forge & Sponge versions.
We should mention somewhere in the docs on what version of forge to use (if we don't already; I couldn't find it) explaining what the filename of the Sponge Forge Mod might reduce the amount of questions and errors that get reported on the forums.

On the Sponge Maven repository, the builds are named according to 
{MCVersion}-{ForgeVersion}-{SpongeAPIVersion}{SpongeBuild}

https://forums.spongepowered.org/t/weird-plugin-error/7027/15

saladoc:
Fixes #194

### DIFF
--- a/source/server/getting-started/implementations/sponge.rst
+++ b/source/server/getting-started/implementations/sponge.rst
@@ -12,6 +12,31 @@ Download
 
 Check :doc:`../../../downloads` for further information.
 
+Choosing the correct version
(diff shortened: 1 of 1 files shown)


boformer:
Nice work!

Issue #208, opened 2015-05-09
Mention CommandException


boformer:
command exceptions can be used to send error messages to the client.

Issue #202, opened 2015-05-07
Text documentation should be placed before Command doc


boformer:
Understanding the text API is a requirement for working with commands

BlackGoat:
Re-ordering is OK, but that should also be mentioned at the top of the commands page (with link).
PS: Thanks for all the scrutiny, it is much appreciated.

Issue #199, opened 2015-05-07
Explain Command Result


boformer:
A command executor returns a result object that is passed to command blocks.

There is a builder to create these command results.

Issue #197, opened 2015-05-06
Add information on how to override / underride vanilla commands


ryantheleach:
And when to register commands.

>https://github.com/SpongePowered/SpongeDocs/pull/183#issuecomment-99098988

If there is an equivalent of a command pre processor for scripting plugins such as skript or CommandHelper it should also be mentioned that it shouldn't be used for general command registration (if possible) but only to intercept commands in general.

boformer:
By the way, the recommended event to register commands should be set to ``InitializationEvent``:

http://docs.spongepowered.org/en/plugin/basics/commands.html#the-command-service

ryantheleach:
Can commands be registered later / dynamically? or is there a hard cut off?

boformer:
There is nothing that would prevent that. It was one of the goals of Sponge to provide a dynamic command system.

ryantheleach:
So registering a command returns an optional commandmapping, I just found out that sponge automatically registers disambiguation aliases with the plugin name as a prefix, so it's extremely unlikely that command registration would return absent. The only situation I can imagine it happening in, is if the same command were somehow registered twice. In which case it's almost certainly an error.

ST-DDT:
@ryantheleach AFAICT the issues that this issue is about are:
* **plugin -> commands -> creating**: Mention during which event the commands should be registered (``GameInitializationEvent``) / add link to ` CommandManager` docs
* **plugin -> commands -> creating**:  Mention disambiguation aliases
* **plugin -> commands**: Mention dynamic command (un)registration
* **server -> management**: Configure precedence of commands with the same alias

Refs: 
* https://docs.spongepowered.org/stable/en/plugin/commands/creating.html
* https://docs.spongepowered.org/stable/en/plugin/commands/service.html

ryantheleach:
It was originally **just** about precedence vs Vanilla, and the effects of the order that commands are registered in. I mainly made this issue because I would have been searching for the information back in 2015, and being unable to find the answer.

It's been a long time since then, so the specific behavior I may have been talking about may have been a bug.

But If I recall correctly it used to be possible to register your command before vanilla did, and that would influence who 'won' the command for players, and that was confusing. At the time I thought it was intended behavior, now I'm not so sure, especially considering the disambiguation aliases.

I also used to make the 'mistake' of not realizing just how early preinit was.

PR #216, opened 2015-05-12
Command Additions

saladoc:
Includes
 - Index reordering (#202)
 - `CommandResult` explanation (#199)
 - `CommandException` explanation (#208)
 - Change preferred phase for command registration (adresses boformers comment on #197)

Removes
 - FLARD


source/plugin/basics/commands.rst
```diff
@@ -340,14 +344,35 @@ Use the ``setChildren()`` method of the parent command builder to apply the chil
 
     If a ``CommandExecutor`` was set for the parent command, it is used as a fallback if the arguments do not match one of the child command aliases.
     Setting an executor is not required.
-    
+
+Command Results
+===============
+
+The ``CommandExecutor::execute()`` method must always return a ``CommandResult``. In most cases it is sufficient to return ``CommandResult.success()`` if the command was successful or ``CommandResult.empty()`` if it wasn't. 
+In cases where more information needs to be conveyed, a ``CommandResult.Builder`` should be used. It provides the methods ``affectedBlocks()``, ``affectedEntities()``, ``affectedItems()``, ``queryResult()`` and ``successCount()`` methods, each accepting an integer and setting the attribute of the same name. All attributes that are not set by the builder will be absent.
```
boformer: CommandResult.builder()...

source/plugin/basics/commands.rst
```diff
@@ -340,14 +344,35 @@ Use the ``setChildren()`` method of the parent command builder to apply the chil
 
     If a ``CommandExecutor`` was set for the parent command, it is used as a fallback if the arguments do not match one of the child command aliases.
     Setting an executor is not required.
-    
+
+Command Results
+===============
+
```
boformer: Maybe a short explanation how the result is used, with link to minecraftwiki commandblock page

boformer:
Maybe we should split up the page (like the block pages)? What do you think?

source/plugin/basics/commands.rst
saladoc: You're right, its simpler to just refer to the `builder()` method. Will change.

source/plugin/basics/commands.rst
saladoc: I'd have to understand it first, so dont expect it tomorrow morning ;)

boformer: http://minecraft.gamepedia.com/Command_block#Output

> A command block can power a redstone comparator facing away from it (possibly separated by a block) with a signal strength specific to the last executed command (for example, the number of players affected by the command).
> The signal strength always reflects the last command executed, even after the command block is deactivated. 

and

> Command blocks can post various success results to scoreboard objectives, including:
> SuccessCount: the number of successes (previously also available by redstone comparator output, but not limited to 15)
> AffectedBlocks: the number of blocks that were altered
> AffectedEntities: the number of entities that were altered (including players)
> AffectedItems: the number of items that were altered
> QueryResult: value returned by command

saladoc:
@boformer I already thought about that since the page has now quite large. I'll look into it.

BlackGoat:
Looks OK, splitting the page was a good idea. Will there be any matching content on the page plugin/advanced/commands/arguments, or will that be dealt with separately?

boformer:
There will not be any matching content on that page. It is a separate issue. 

We could merge the two "arguments" pages, so that only the CommandCallable page remains in the advanced section.

BlackGoat:
Thanks. Do whatever makes the most sense to you. Someone let me know when this is ready to merge.

boformer:
Except for the last commit, this is ready.

saladoc:
That's why its the last commit, so it can manually be left out. My reasoning behind already including it was that 
 - probably most developers use the latest API snapshot anyway
 - the change is fully implemented in the API, so it will be available in any Sponge implementation that uses a SpongeAPI version newer than the linked commit. Which would be any current dev build.
 - if all docs had to be strictly 2.0, I'd have to change part of the block api docs since the `ImmutableDataHolder#withoutData()` method was introduced in SpongePowered/SpongeAPI@31467e86 - a 2.1 snapshot version.

As far as I am concerned, this can be merged. If strict compliance with SpongeAPI 2.0 is intended, just leave out the last commit.

BlackGoat:
The dependencies page now demands Sponge API version 2.1-snapshot, so the rest of the docs must follow suit. Blame https://github.com/SpongePowered/SpongeDocs/commit/922ce29ae9336aa186a3730566779c81d8955b27

PR #219, opened 2015-05-13
Injection Update

FerusGrim:
Moved injection.rst from 'advanced' to 'basic'. It's an important part of the API, and not all that advanced.

Fleshed out a few of the different types of injections.

I tried to give as many examples as possible, without just repeating the same line of code for each Object.


source/plugin/basics/injection.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Injection Examples
+==================
+
+There are a few references which are difficult to get - or, in some cases, impossible - without injection. While these
+may not be absolutely vital to every plugin, they're quite frequently used.
+
+.. tip::
+
+    Note that it's *almost always* recommended to Inject your objects within the main class, as it's instantiated with the Guice injector when the plugin is loaded.
```
BlackGoat: Shouldn't Inject be inject (ie. no caps)? PS: This tip is also a Note :)

source/plugin/basics/injection.rst
```diff
@@ -0,0 +1,178 @@
...
+    @Inject
+    public Apple(Game game) {
+        this.game = game;
+    }
+
+Config Directory
+~~~~~~~~~~~~~~~~
+
+This is a big one, especially if you plan on having more than a single file stored for your plugin to later call.
+Thankfully, with the power of the Guice injector, it's never been easier to get your directory.
```
gratimax: We're not trying to advertise here, we've already sold people on "how easy it is" to make Sponge plugins... The goal is to describe features in an unbiased tone

FerusGrim: If anything, I was "advertising" the Guice injector. It's damn awesome. And why wouldn't we want to inform people of "how easy it is" to make Sponge plugins? Shouldn't that be exactly what we're doing?

We're not journalists, we're literally the software we're propping up. I don't envy your opinion on how commercials should sound. :p

source/plugin/basics/injection.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Overview
+========
+
+Dependency injection allows plugins to designate a few API types that will be injected after construction.
+
+Temporary List of Injected Types
+================================
+
+* Game
```
boformer: Maybe use a definition list: http://docutils.sourceforge.net/docs/user/rst/quickref.html#definition-lists

FerusGrim: I'll leave that up to @BlackGoat. I didn't actually edit that part - it just looks like it because I moved the file.

BlackGoat: Definition list is definitely a good idea. Modify it if you want.

PR #259, opened 2015-06-07
Fix note in pagination

ThinkChaos:
See bottom of the [pagination](https://docs.spongepowered.org/en/plugin/basics/text/pagination.html) docs.


RobertHerhold:
Good catch.

PR #268, opened 2015-06-11
Correct keyboard mistake

Piratdu52:
A ',' between '"', instead of next the last.

### DIFF
--- a/source/contributing/docs/index.rst
+++ b/source/contributing/docs/index.rst
@@ -5,7 +5,7 @@ SpongeDocs Writing
 Overview
 ========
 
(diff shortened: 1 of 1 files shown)


kenzierocks:
Punctuation is always inside quotes.

RobertHerhold:
I would previously have agreed with @Piratdu52, but I looked it up [here](http://www.grammarbook.com/punctuation/quotes.asp) and @kenzierocks is correct :)

From the site: `Rule 3a. Periods and commas always go inside quotation marks.`

Piratdu52:
Sorry, it's not the case in my language. I will delete this fork ...

PR #273, opened 2015-06-14
Normalize faq.rst by the SpongeDocs style guide.

Zarthus:
Fixes some inconsistent casing, remove a duplicate heading, and reword some titles to make more sense.

### DIFF
--- a/source/about/faq.rst
+++ b/source/about/faq.rst
@@ -5,11 +5,8 @@ Frequently Asked Questions
 For Everyone
 ============
 
-Can I use SpongeAPI to create client side plugins?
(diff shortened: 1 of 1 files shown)


RobertHerhold:
I think "Can I Create Client Side Mods With Sponge?" would suffice.

RobertHerhold:
Also I think "Are There Downloads Available?" could be better stated as "Are Downloads Available?".

RobertHerhold:
Whoops! I meant "Partitioned" as an actual substitution for "Broken Down". I'll fix it when I get back from vacation.

Zarthus:
ah, I see now. It's not very significant luckily. ^^

RobertHerhold:
@Zarthus It's fine; I just fixed it

PR #275, opened 2015-06-16
Maintain consistency throughout the staff page

jamierocks:
If someone is the leader of multiple things: Community and IRC Leader.

If there are two unrelated things such as Web Leader and Developer - they should be on separate lines.

Also 'and' should be used and not '&'.


RobertHerhold:
By that logic, why wouldn't Zidane's roles as "API Leader" and "LLC Director" be on two separate lines

jamierocks:
Yes, he would. :P Fixing that now

thanks @RobertHerhold :)

RobertHerhold:
Thank you :)

jamierocks:
No problem :)

PR #281, opened 2015-07-03
fix mentions of mcpc+/DMCA in history.rst

Tzky:


necrodoom:
The reasons i believe these changes are necessary:

1. Cauldron is a successor (or a rename of) MCPC+, last MCPC+ version is of an year before DMCA, and so irrelevant to the topic.

2. Remove DMCA mention as there were many factors in bukkit's death, including but not limited to:

2a. Bukkit team annoucing a development halt

2b. Bukkit team quitting after Mojang's annoucement and silent removal of Evilseph

2c. Lock down of bukkit repos by Mojang employees

Tzky:
Edited the controversional passage for the reasons mentioned above. 
Thank you Necrodoom for pointing this out!

PR #282, opened 2015-07-03
The nitpick of all nitpicks

necrodoom:


simon816:
IMO SpongeDocs should have a minor issues list like SpongeAPI and SpongeCommon does. That way these very small changes can be fixed all in one go.

PR #296, opened 2015-07-15
Update blocks documentation to match API

12AwsomeMan34:


BlackGoat:
Nice job, an update *and* some typographical errors fixed. The funky names for code symbols are out of our control here @boformer , and it seems otherwise OK. Happy to merge this if you are :)

boformer:
Why are the parameter names out of our control?

BlackGoat:
If it's an arbitrary internal parameter, go crazy, I thought you were nit-picking the API. My bad.

PR #297, opened 2015-07-15
Update on entity docs

12AwsomeMan34:


source/plugin/basics/entities/spawning.rst
```diff
@@ -11,22 +11,24 @@ For example, let's try to spawn a Creeper:
...
+        Optional<Entity> optional = extent.createEntity(EntityTypes.CREEPER, location.getPosition());
         if (optional.isPresent()) {
             // After this, we can use more API that relates to creeper
             Creeper creeper = (Creeper) optional.get();
-            creeper.setPowered(true);
-            creeper.setExplosionRadius(10);
+            ExplosiveRadiusData radiusData = creeper.getExplosiveRadiusData();
+            radiusData.setExplosionRadius(10);
+            // Return the modified explosive radius data back to the creeper
+            creeper.offer(radiusData);
```
gabizou: Why not add `ChargedData` and `DisplayNameData` to show a bit more of an example?

gabizou: Also, given that this is all `Data API`, we don't even need to cast anymore to `Creeper`.

gabizou: @12AwsomeMan34 pinging for changes requested.

12AwsomeMan34: Will get to doing this after work.

Issue #298, opened 2015-07-18
Persistent Typo & Nitpick List - for small corrections.


Inscrutable:
This Issue is intended to be a List collecting minor typographical errors and small nitpicks together, so they can be corrected collectively - and to stop little bugs from going unreported. It should be checked (and hopefully cleared) regularly. Some issues here might be bundled and fixed in other PRs.

Here is the list: (Cleaned up 16-8-15)
- [x] Page: sponge commands: add `/sponge help` command
- [x] Page: plugins - blocks: rename all Location parameters from *block* to *blockLoc*
    - moved to #263 
- [x] Page: plugin/main-class - "Plugins are loaded before the game and the world." -> world(s)
- [x] Page: server/management/permissions should refer to list on server/spongineer/commands

jamierocks:
Extra space in SpongeAPI
https://github.com/SpongePowered/SpongeDocs/blob/master/source/index.rst

12AwsomeMan34:
As mentioned in #296, it may be beneficial to rename all `Location` parameters from `block` to `blockLoc` in the blocks documentation.

ryantheleach:
[] https://github.com/SpongePowered/SpongeCommon/commit/6fc98451fafd784f0b3b4c7cf342b785a1fc056c#diff-f5830811c15e1734b3a69e5658f1ea85R81 Added a help command.

Inscrutable:
What's left of this is moved to #263

PR #299, opened 2015-07-18
Fix reference to Mixin

jamierocks:


BlackGoat:
Well spotted. Can you please add minor issues like this to #298 in future?

PR #300, opened 2015-07-18
Changed references from Forge to MCP

jamierocks:


Minecrell:
It is kind of correct as-is, because NMS is completely unsupported on SpongeVanilla currently.

jamierocks:
Oh?

PR #301, opened 2015-07-18
Remove unnecessary space in SpongeAPI

jamierocks:


RobertHerhold:
Please use #298, as @BlackGoat mentioned in #299

jamierocks:
Too much effort :/

RobertHerhold:
Sorry, but that was the decision that was made. Please bring it up with @BlackGoat if you would like to discuss it.

BlackGoat:
There is no real rule here, more of a guideline. If it's already a PR, I'd just let it slide. I'm not that fussy. The point is, it should be easier to post them as comments, and let them all be wrapped up at once.

PR #307, opened 2015-08-03
Add guideline for maximum character line length

RobertHerhold:
This would be purely for the ease of writing the docs. It gets rather annoying to be scrolling side to side, so I think we should set a 120 (plus/minus) character line limit. What's the opinion on this?

@BlackGoat


Spongy:

BlackGoat:
RtD seems to be lacking any variable screen width options, so a maximum width is a good idea. Enforcing it for Tables may be more difficult. Code blocks might also violate this here and there. Perhaps they may be allowed exceptions?

RobertHerhold:
I wouldn't mind for a few things like tables/code blocks, but I think in general there should be a general upper limit

Tzky:
80 chars is the old limit which is rather low. I'd go with a number which fits to our docs theme. I suggest something like 100 to 120 chars.

Oh and i'd volunteer editing the docs to fit this rule.

RobertHerhold:
@Tzky 
Thanks for offering, but I am pretty much finished with the reformatting

RobertHerhold:
#308

PR #313, opened 2015-08-07
Spice up the FAQ.

Tzky:
This PR:
* adds a nice and shiny table at the top of the FAQ
* changes some Q's for better readability

As always comments and thoughts are more than welcome :)

### DIFF
--- a/source/about/faq.rst
+++ b/source/about/faq.rst
@@ -23,7 +23,7 @@ Sponge (and Minecraft) needs the Java Runtime Environment to run properly. You w
 the server on too, besides that nothing is required. Learn more about choosing the correct Java version here:
 :doc:`../server/getting-started/jre/`
 
-Where do i get Sponge?
+Where do I get Sponge?
 ----------------------
 
 Sponge currently isn't finished but there are development builds of the SpongeCoremod available: :doc:`/downloads`
@@ -41,11 +41,11 @@ The Sponge Project currently develops two implementations which both use the Spo
 (2) **SpongeVanilla**, a stand-alone implementation of the Sponge API, running on top of the vanilla Minecraft server.
     (SpongeVanilla was formerly known as Granite, until the development teams merged).
 
-Where do i get Plugins for Sponge?
+Where do I get Plugins for Sponge?
 ----------------------------------
 
-Yes, and it is called Ore. Development has already started on GitHub. Temporarily plugins are hosted on the
-SpongeForums.
+On our official plugin repository, which is called Ore. Development has already started on GitHub. Temporarily, plugins
+are hosted on the SpongeForums.
 
 For Server Owners
 =================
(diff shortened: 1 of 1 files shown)


source/about/faq.rst
```diff
@@ -2,89 +2,56 @@
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+(2) **SpongeVanilla**, a stand-alone implementation of the Sponge API, running on top of the vanilla Minecraft server.
+    (SpongeVanilla was formerly known as Granite, until the development teams merged).
 
-The developers of Granite have joined the Sponge project, and Granite was renamed SpongeVanilla. It is now an officially
-supported Sponge API implementation, developed in parallel with Sponge coremod.
+Where do i get Plugins for Sponge?
+----------------------------------
 
+Yes, and it is called Ore. Development has already started on GitHub. Temporarily plugins are hosted on the
```
12AwsomeMan34: "Where do i get Plugins for Sponge?"
i -> I

And the response should not be "Yes", for a where question.
Temporarily should have a comma after it, to signify a pause.

source/about/faq.rst
```diff
@@ -110,8 +77,21 @@ Forge installed - unless your server runs Forge mods, of course.
 For Developers
 ==============
 
+What can i do with Sponge?
```
RobertHerhold: i -> I

source/about/faq.rst
```diff
@@ -110,8 +77,21 @@ Forge installed - unless your server runs Forge mods, of course.
...
+--------------------------
+
+Sponge provides a PluginAPI. This means that you can create new content and gamemodes on the go.
```
RobertHerhold: PluginAPI -> Plugin API (PluginAPI is not a proper Sponge term)

source/about/faq.rst
```diff
@@ -110,8 +77,21 @@ Forge installed - unless your server runs Forge mods, of course.
...
+
+
+What can't i do with Sponge? / Limitations of Sponge?
```
RobertHerhold: i -> I

source/about/faq.rst
```diff
@@ -110,8 +77,21 @@ Forge installed - unless your server runs Forge mods, of course.
...
+-----------------------------------------------------
+
+Sponge can't be used to create new blocks, textures, mobs or any other content which would need clientside modifications.
```
RobertHerhold: clientside -> client-side

source/about/faq.rst
```diff
@@ -110,8 +77,21 @@ Forge installed - unless your server runs Forge mods, of course.
...
+However you can make use of the ForgeAPI for clients and create Sponge plugins for the serverside.
+It is even possible to use Sponge on the clientside, but for several tasks mods are still required.
```
RobertHerhold: clientside -> client-side

Tzky:
While i'm glad your're pointing this faults out, i'd be even happier if you could comment on the questions asked and if you like the idea of having a link list at the top or not. Example:

unbedsgsdfdsfsdfanannt

RobertHerhold:
@Tzky 
My apologies; I assumed this PR was up for review and not just a what-do-you-think PR. I like the idea, but this brings up this issue: Do we do this for every page? Do we have a synopsis/table of contents at the top of each doc?

BlackGoat:
The versions question(s) is a bit too long, perhaps just "What Sponge implementations are available?"
The "For Developers" section might be better with one "What can I do" page that links directly to the Plugins index page (and the Bukkit question, it's *still* frequently asked). There's three rogue Plugin Dev questions in the Server Owners bit too, that might get rolled in.
I guess we need to make sure the rest of these questions *are* frequently asked. I'm sure the server-internals one has come up a few times recently too.

Tzky:
@RobertHerhold No offense made. I just didn't state that clearly enough, my apologies too.

To answer your question: I don't want to use this on every page as the docs would get bloated fast. I thought of this as a quick reference and to avoid having to scroll the complete faq before you know the content of it.

Regarding the content itself: I'm not sure which questions should get added/removed/modified. If your think something eagerly needs an edit, just tell me as Inscrutable did above.

Tzky:
Closed this to move my changes from my fork to the SpongeDocs repo. This gets us the nice and shiny Sponge preview.

PR #314, opened 2015-08-08
Refactor migrating pages

Tzky:
* merged migrating pages
* reduced nestes pages
* added overview to new migrating page

As always comments on the concept as well as pointing faults out to me is appreciated!
(Hopefully i can utilize Spongie this time...) -> at least this seems to work.


Spongy:

BlackGoat:
I think it would look better with horizontal bars separating the different sub-sections of *Migrating*. The sections need to stand apart for readability. The TOC is very long, too. I wonder if it would be better to have one less TOC level.

Tzky:
I tried both versions, the TOC with less depth looks better. I also introduced horizontal lines between the subsections, but i'm unsure of this looks better.

BlackGoat:
This looks correct to me. Changes will come later with SpongeForge, and whenever SpongeVanilla is released (Migrating to vanilla would be almost identical to SpongeForge anyway). Squash away.

PR #308, opened 2015-08-04
Reformat to 120 character line limit

RobertHerhold:
Phew... that took a long time. I wouldn't bother reading through the diff, but feel free to browse through Spongie's preview for any errors.

This should be pulled in along with #307

PR #316, opened 2015-08-09
Condense contributing section

Tzky:
This PR cleans up the mess in /contributing/ leaving just a few pages
in.

This is achieved by removing old (empty) files, merging SpongeForge and
Vanilla into a single implementation page and replacing the Docs files
with a single page.

Comments and input on this is appreciated.


Spongy:

Tzky:
@gabizou will write up some docs on how to contribute to SpongeCommon. He suggested that his write-up should get pulled into the docs when done. This will probably affect this PR and the structure of the docs too.

Link: https://github.com/SpongePowered/SpongeCommon/wiki

BlackGoat:
What happened to the 120-character limit guideline? #307 seems to be reverted somewhere... Better check #308 too. Something odd is going on here.

Tzky:
Should be fixed now. I also added a temporary link to the SpongeCommon wiki, as long as it's not finished. Afaik we will pull that one in when it's done.

This should be ready for merging, too.

RobertHerhold:
Looks good. Thanks for cleaning this up :)

Issue #324, opened 2015-08-31
Update SpongeDocs to Java8


Tzky:
With the recent announcement that Sponge will (finally) move to Java8 and drop support for 6 and 7 the Docs now must get updated to fit Sponge.

reference: 
https://forums.spongepowered.org/t/theres-a-java-update-available/9147
PR adressing this Issue: #325

Minecrell:
We should also extent the docs to provide more help with installing Java 8, especially on quite frequently used distributions like Ubuntu Trusty (14.04 LTS) where no official openjdk8 build is available yet. It would be good if we have a clear explanation how to update for users not that experienced with this kind of stuff.

boformer:
That means we have to redo every code snippet which contains ``Optional``s...

gabizou:
> That means we have to redo every code snippet which contains Optionals...

Sounds about right. But it's just an import change and `fromNullable` to `ofNullable` and `orNull` to `orElse(null)`.

Inscrutable:
Only two pages need to change java version (J6/7 -> J8): server/getting-started, & preparing/jdk.
The rest seems to involve reviewing code snippets for ``Optional``.
(We should make sure this isn't merged until the master branches are similarly tweaked.)

Tzky:
I started working on this in java8 branch. Feel free to push your changes there :)

PR #325, opened 2015-09-01
Update SpongeDocs to Java8

Tzky:
With the recent announcement that Sponge will (finally) move to Java8 and drop support for 6 and 7 the Docs now must get updated to fit Sponge.

ToDo-list i can currently think of:
- [x] check all code snippets for validity, modify where needed
 - [x] use Java Optionals instead of Guava
 - [x] remove import of Guava
- [x] state that Sponge needs Java8 to run
- [x] modify installation insctructions for JRE and JDK to require Java 8
- [ ] add instructions for Linux, where Java 8 sometimes isn't easily available

reference: 
https://forums.spongepowered.org/t/theres-a-java-update-available/9147
Issue:  #324

### DIFF
--- a/source/contributing/guidelines.rst
+++ b/source/contributing/guidelines.rst
@@ -167,7 +167,7 @@ started quickly, here is an example of properly formatted code:
             if (random.nextBoolean()) {
                 return Optional.of(this.id);
             } else {
-                return Optional.absent();
+                return Optional.empty();
             }
         }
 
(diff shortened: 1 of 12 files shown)


Spongy:

saladoc:
I've checked all code examples to be found within the plugins section.

Most don't even bother to list the `Optional` import so those are fine. I changed those who did import it and fixed other small errors where I spotted them.

Tzky:
Alright, seems only the Linux instructions are missing to get this finished. At least for now...

Inscrutable:
On the MC Wiki page [Setting up a server](http://minecraft.gamepedia.com/Tutorials/Setting_up_a_server) there is a section for Linux instructions, including getting Java 6 or 7 for many distros. This information could be adapted for our purposes, substituting 1.8 java versions.

saladoc:
In the java 8 `Optional`, the method `absent()` does not exist and needs to be replaced with `empty()` Example occurence in `plugin/commands/arguments`:
```
Optional and weak arguments may return ``Optional.absent()``.
```
needs to be
```
Optional and weak arguments may return ``Optional.empty()``.
```

PR #345, opened 2015-10-19
Add rule about evasion

Phroa:


saladoc:
Rule change approved by @Zidane in private conversation.

PR #360, opened 2015-11-02
Make TOC tree smaller on index

gratimax:
Make TOC tree smaller


Spongy:

Tzky:
I think this is the best possible solution to de-bloat the index page. 
I'd use this and leave the rest of the page alone.

RobertHerhold:
+1

PR #362, opened 2015-11-04
[WIP] remove JD hardlinking. fixes #127

### LINKED ISSUE #127: Implement a way of adding javadoc links without hard-coding the URL
Currently, the only semi-official javadoc homepage is marked as "temporary link", yet occasionally it is important for the documentation to provide a link to a certain javadoc.

The preferred way to do so would be a custom rst role like
```
for details about how Sponge will soak up FLARD see :javadoc:`org.spongepowered.api.soaker.FLARDSoaker`
```
or
```
the first step to not being seen is to call the :javadoc:`hiding.Nesbit#setNotStandingUp` method
```

Thoughts:
 * default link text is the name of the class / method referenced
 * base url for javadocs is configured while building the docs
 * no more hard-coded javadoc links in the documentation

According to http://sphinx-doc.org/extensions.html#where-to-put-your-own-extensions it is possible to provide project-local sphinx extensions. I think that may be a legit way to solve this. 
There is also a built-in extension to include external links to source code which maybe can be abused to achieve this: http://sphinx-doc.org/ext/linkcode.html

Tzky:
This PR adds 'sphinx.ext.extlinks' module to conf.py. This allows us to add non-hardcoded javadoc links to the docs: `:javadocs:'link goes here'`

After further investigation i think that the javalink package for Sphinx is the way to go. This however requires adjustment to the current build process:
- [ ] update Sphinx to 1.3
- [ ] install sphinx-javalink + dependencies
- [ ] find a way to retrieve *.jar for the required classpath variable

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -27,7 +27,11 @@ import sponge_docs_theme
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = []
+extensions = ['sphinx.ext.extlinks']
+
+extlinks = {
+'jdpackage': ('https://jd.spongepowered.org/index.html?%s/package-summary.html',' '),
+'jd': ('https://jd.spongepowered.org/index.html?%s.html',' ')}
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
--- a/source/index.rst
+++ b/source/index.rst
@@ -5,6 +5,12 @@ Sponge Documentation
 Introduction
 ============
 
+Package: :jdpackage:`org.spongepowered.data.manipulator <org/spongepowered/api/data/manipulator>`
+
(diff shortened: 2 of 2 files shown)


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/8eacab25/index.html.

- modified https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/8eacab25/conf.html
- modified https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/8eacab25/index.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/362#issuecomment-153671631), not to the direct url to the preview._

Tzky:
After further investigation, this way doesn't work nicely when linking methods. At least not out of the box.

The second option (javalink) is built against python 2.7 (iirc we use 3.x) and requires jar files of all desired jd repos to be present while compiling. I think that's not an option either.

Will close this PR for now until there's a better option available.

Issue #367, opened 2015-11-06
Index TOC in 2 columns


Inscrutable:
I believe we should try to do something to group the links on the Main and Plugin Index pages, to assist in readability. [continued from #366 ] Two columns (with Toctrees inside a table) seems to be possible, thanks @Tzky. Let's work on *How* we reorganise *what*. :)

Tzky:
Example code how this could work: https://gist.github.com/Tzky/f4d072de5d6d06a4421f
Preview: 
![toccolumn2](https://cloud.githubusercontent.com/assets/10940670/10996946/667db7a6-8489-11e5-9755-5623de08677c.png)

Tzky:
I'll close this for now as i closed the PR too.

PR #372, opened 2015-11-12
[ON HOLD] Refine TOC, addresses #367

Tzky:
This PR condenses the index pages and especially the toctrees even further.
It makes use of the suggestion i made in #367.

My question is:
Does this look better than before?


Spongy:

source/index.rst
```diff
@@ -1,92 +1,86 @@
...
-
-* Developers who wish to create Sponge plugins or contribute code to the Sponge project
-* Users who wish to help with writing and editing articles in SpongeDocs
-
-Please read this section before `Creating Plugins`_ and `Contributing to Sponge`_.
-
-.. toctree::
-    :maxdepth: 2
-    :titlesonly:
+This section is covers the basic setup (and how to do it) of your workspace wether you want to develop for Sponge your
```
phroa: is covers; wether; "develop for Sponge your own plugins" -> "develop plugins for Sponge"

source/index.rst
```diff
@@ -1,92 +1,86 @@
...
+|    :maxdepth: 2    |  |    :maxdepth: 2       |
+|    :titlesonly:    |  |    :titlesonly:       |
+|                    |  |                       |
+|    plugin/index    |  |    downloads          |
+|                    |  |                       |
++--------------------+--+-----------------------+
+
+
+
+The Sections in Detail
```
phroa: I suggest "Section Summaries"

source/index.rst
phroa: I would also remove the bit in parens.

Inscrutable:
I like the restructuring of the Plugin index, all on one screen. I'm not so happy with the main index, because the text is separated from the links. Maybe you can place the text in the table instead, so that links are parallel to their related text?

source/index.rst
Tzky: ah, the phrasing... ``If you want to develop for Sponge, your own plugins or help expanding the docs``.

The bit in parens? I'm sorry, but i don't get it.

Tzky:
I made 5 different versions to show off. What's *your* preference?

Inscrutable:
It is a great pity the TOC wraps horribly on (1), because that would have been my preferred format.
Version (3) seems most readable to me. I dislike the text as footnotes in (4), and (5) is too spartan. 
I will have a go at rewriting (and minimising) the text when the format is decided. Any more opinions?

gratimax:
Doesn't look good on mobile devices. The previous versions didn't really either, but this makes it nearly unreadable.

saladoc:
Version 1 looks just plain weird. Version 2 I deem very confusing, version 3 being only slightly better.
Version 4 I would like the most if it weren't for the strange ordering and the greater distance between a sections links and its description.

I have come to think that the best way to reduce the clutter on the index page is to further lower the `:maxdepth:` to 1, thus having only the link to a sections index page along with a summary of that section.

Tzky:
Gratimax, which version looks weird on mobile devices? Could you explain a bit further?

progwml6:
Those who are having UI issues on mobile: please provide a screenshot along with the type of device & browser used.

gabizou:
Just posting for informational purposes. 

Using an iPhone 6+ with iOS 9
http://imgur.com/gallery/phGUn

Tzky:
Android 4.4.4 on Xperia Z1 Compact, stock browser (Chrome):
http://imgur.com/a/q6Gbm

Tzky:
I'll close this PR as i don't think this change is what we're currently heading for. We'll probably update the TOCtrees when we push the unified design. For now we'll leave it as it is. I'll port all other changes to another PR.

PR #378, opened 2015-11-24
Add GAME_STOPPING and GAME_STOPPED.

### LINKED ISSUE #376: GameState.GAME_STOPPING and GameState.GAME_STOPPED
Add GameState.GAME_STOPPING and GameState.GAME_STOPPED to the Plugin Lifecycle page

https://github.com/SpongePowered/SpongeAPI/commit/14752984c69795ca609fe09d559632d552b349fb

JBYoshi:
Closes #376.

### DIFF
--- a/source/plugin/lifecycle.rst
+++ b/source/plugin/lifecycle.rst
@@ -85,8 +85,10 @@ During this state, no players are connected and no changes to worlds are saved.
 Stopping States
 ~~~~~~~~~~~~~~~
 
-Stopping states occur only once during a single run. ``SERVER_ABOUT_TO_START`` may follow ``SERVER_STOPPED``,
-and ``SERVER_STOPPED`` may occur at any point during the process if there is an error.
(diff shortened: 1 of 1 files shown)


saladoc:
Thanks for the docs (and implementation) for those events. When I spoke to Zidane about it he mentioned that the events may not be guaranteed to fire in some cases (Server gets ctrl-c'd instead of propertly `stop`ped, Client gets closed using the X button instead of "Quit Game". Since you surely are testing your SpongeForge PR, can you add a `.. warning::` detailing in which cases the shutdown events won't fire?

Also I'd be glad if you were to include some best practises for the given events (Plugins exposing an API should still be able to accept requests during `GAME_STOPPING`, in `GAME_STOPPED` no more interaction between plugins should occur).

Deamon5550:
Some testing should be done to find out when these are called and when they aren't, for instance SIGTERM or SIGINT (such as from ctrl-C) will still cause an orderly shutdown.

PR #382, opened 2015-12-06
Updated Scheduler for new Builder system, and corrected-expanded delay/interval methods

theboomer:
Updated scheduler to reflect the newly refactored package location, and
Task.Builder object

Modified Delay and Interval properties to show BOTH methods and replace
no longer accurate "..optional TimeUnits, or else defaults to Ticks..".

Removed information about async use and defaults in absence of
parameters, because it was not clear, and I dont know what of it is
actually correct in the new system.  Someone please re-add asynch
comments for delay/interval descriptions if they are applicable.

### DIFF
--- a/source/plugin/scheduler.rst
+++ b/source/plugin/scheduler.rst
@@ -7,7 +7,7 @@ Sponge exposes the ``Scheduler`` to allow you to designate tasks to be executed 
 name, (a)synchronicity, and ``Runnable`` (see :ref:`task-properties`).
 
 Working With the Task.Builder
-============================
+=============================
 
 First, obtain an instance of the ``Scheduler``, and retrieve the ``Task.Builder``:
 
--- a/source/plugin/scheduler.rst
+++ b/source/plugin/scheduler.rst
@@ -2,25 +2,25 @@
 Using the Scheduler
 ===================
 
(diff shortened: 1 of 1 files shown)


source/plugin/scheduler.rst
```diff
@@ -67,24 +71,24 @@ Using the ``TaskBuilder``, you can specify other, optional properties, as descri
...
-	SchedulerService scheduler = game.getScheduler();
-	TaskBuilder taskBuilder = scheduler.createTaskBuilder();
-
-	Task task = taskBuilder.execute(new Runnable() {
-		public void run() {
-			logger.info("Yay! Schedulers!");
-		}
-	}).async().delay(100, TimeUnit.MILLISECONDS).interval(5, TimeUnit.MINUTES)
-        .name("ExamplePlugin - Fetch Stats from Database").submit(plugin);
+    Task task = taskBuilder.execute(new Runnable() {
```
simon816: Could convert this to a lambda expression as it's neater. Also could show that you can use a `Consumer<Task>` here instead.

theboomer: I was simply replacing the existing system with corresponding changes.
I'm not confident enough with the java8 features to make the change leap either, perhaps this could be done afterwards, or as a secondary alternative approach to show how they bridge the same solution.

PR #383, opened 2015-12-07
Fix dead JavaDocs links

Milchshakee:


Tzky:
Nice catch, thanks for reporting that one! :)

PR #384, opened 2015-12-07
Streamline the index pages (again...)

Tzky:
* remove unneccessary headings on index pages
* some rephrasing
* moved/merged some pages


Spongy:

Inscrutable:
This looks OK to me. It removes some of the original "friendliness" of the pages, but now there is more significant content there anyway. Streamlining is always good, as long as parts don't fall off.

Tzky:
The reason i removed the headings is that i felt that they're unneeded and tend to just bloat the pages without further benefit. This can make the pages look less friendly and a bit more functional, which is why i wanted to get feedback first :)

PR #386, opened 2015-12-09
enable favicon in conf.py

Tzky:
Adds a favicon to the SpongeDocs. This must be pulled in together with or after SpongePowered/Sponge_docs_theme#9 has been merged.

Note that the build will fail as long as the theme PR isn't merged.


Spongy:

gratimax:
:+1:

PR #400, opened 2015-12-26
Fix Error in Data Docs

saladoc:
- Remove false information
- add clarification about removing data / meaning of absent data

@Aaron1011 I tried to work your information from #398 into this, however I'm not sure about `DisplayData` always being present on players. I have taken a look at the corresponding data processor and it seems to not treat players any different from other entities.


Spongy:

PR #407, opened 2016-01-01
Update the Docs to 3.0.0 and Beta

Tzky:
* Bumps the Docs version to target API 3.0.0 release
* Updates the Docs to reflect Beta release

This PR updated the Docs to reflect the recent version change. Afterwards we will need to introduce a new branching/versioning model on the Docs (just like the API did), but i'd like to keep it KISS as the Docs are a somewhat rollingrelease anyways.

Also i'd like to keep the bump ver commit and the Beta changes commit separated.


Spongy:

boformer:
Good.

source/about/history.rst
```diff
@@ -47,7 +47,7 @@ shared code between the two implementations while implementation-specific code r
 On April 21st, 2015 SpongeAPI 2.0 was finally released. A week later development builds of SpongeForge were
 finally made available.
 
-
+On December 31st, 2015 SpongeAPI 3.0.0 was released, along with it SpongeForge entered its beta.
```
kashike: This sounds weird.

source/downloads.rst
```diff
@@ -12,11 +12,14 @@ Download
 SpongeForge
 ~~~~~~~~~~~
 
-* **Status:** Alpha. The first alpha build of SpongeForge (#428, Inspired Wallaroo) was released on April 25th, 2015.
+* **Status:** Beta.
+
+* The first alpha build of SpongeForge (#428, Inspired Wallaroo) was released on April 25th, 2015.
+* The first beta build of SpongeForge (#1000) was released on December 31st, 2015.
```
kashike: No name? :crying_cat_face:

Tzky: I don't think a name was chosen yet ;)

saladoc:
Not good enough.

We find ourselves in the awkward position where we still have lots of catching up to do, yet as development of Sponge continues, we may have to accept changes into the docs that come up with the ongoing development, but will not yet have to be included in the current docs. Up until now, this has not been a problem as we documented a rolling-release api with rolling-release docs.

As of version 3.0.0, Sponge has dropped the rolling release model and the docs must follow suit. However unlike Sponge there is not a single docs commit for any release version as there may be changes in wording or structure that do not affect the content. My proposal for the docs is the following:

* The docs at docs.spongepowered.org must always correspond to the latest released version.
therefore
* The `master` branch contains all changes pertaining to the latest release
* feature branches contain additions and modifications, those feature branches may be aggregated in either an `upcoming-release` branch or a version specific branch

Now if a new sponge version is released, the following happens:

1. the latest commit on `master` is tagged with the version, i.e. `git tag v3.0.0`. 
2. All applicable feature or `upcoming-version` branches are merged into master
3. The version number in the `conf.py` is updated

Docs writing then continues based on the version update commit. It may be the foundation for a new `upcoming-release` branch or for any feature branches.

However, this system would require slightly more attention when merging PRs as it must be carefully determined whether the PR targets the current release (should be merged into `master`) or the upcoming release (targets the `upcoming-release` branch).

boformer:
And even worse, commits which target the current version must also be merged into the next-version branches.

Tzky:
Next-version branches require constant rebasing anyways...

With the new SpongeAPI versioning and branching model, i suggest that we base the Docs brancing model on this *or* we use a KISS based model.

Tzky:
I think this PR is good to go. The decision on the branching/versioning is still pending. I created an issue ( #414 ) to discuss this.

PR #409, opened 2016-01-01
Fix line wrapping issues

12AwsomeMan34:
Fixes the line wrapping issues that were present in the item docs.


saladoc:
Had to make another addition + rebase it on master, thus no fancy "merged" status.

PR #412, opened 2016-01-02
Update index.rst

kamcio96:
Java 8


saladoc:
Thanks for reporting. Please do not start pull requests for small changes. Instead, let us know about the problem either through an issue or a message on IRC. We'll then incorporate the change into one of the next commits.

Issue #402, opened 2015-12-28
Document new exploit patches


Inscrutable:
As of SpongeForge build 974, there are methods included to disable a number of client-server exploits. These are enabled by default, and a verbose logging option is also available. Exploit detection results in the offender being kicked (eg "You have been kicked for attempting to perform a sign command exploit"). The logging can be enabled/disabled via console commands (or via command from an OP). Here are the ones we tested:
 
- sponge config -g logging.exploit-sign-command-updates true
- sponge config -g logging.exploit-respawn-invisibility true
- sponge config -g logging.exploit-itemstack-name-overflow true

There are a number of other minor exploits being worked on, so this list is likely to expand somewhat in the near future. Some of these exploits only work on early 1.8.X versions, due to changes by Mojang.

Flibio:
I could write a page about the exploits, if you would like.

12AwsomeMan34:
Resolved with 45fdf94cbafc5a175b139e93c0069a8b784b2b56 and aa55eda851c7f2b8936f7d6c45d12883f1ccdc38.

PR #413, opened 2016-01-03
Update server configuration docs

12AwsomeMan34:
Updates the server configuration docs.

Current changes:
- /sponge conf -> /sponge config
- Getting the value from a config
- saving and reloading world configs
- Update config example to build 1000

WIP list:
- [x] Update 'Global Properties of Sponge'
- [x] Update commands page to match /sponge config

and any other references to server configuration that is outdated.

PR #418, opened 2016-01-03
Exploit Patch Documentation

Flibio:
Documents how Sponge patches various exploits in Vanilla Minecraft. Closes issue #402.

Depends on #413 to update the global.conf page

---

[Main Preview](http://flibio.net/preview/exploit-patches/)

[File Changed](http://flibio.net/preview/exploit-patches/server/management/exploit-patches/)

### DIFF
--- a/source/server/management/exploit-patches.rst
+++ b/source/server/management/exploit-patches.rst
@@ -1,0 +1,30 @@
+===============
+Exploit Patches
+===============
+
+In recent Sponge builds (SpongeForge 974+), SpongeForge and SpongeVanilla patch a few client-server exploits. Whenever
+the implementations detect a user performing an exploit, they are kicked from the server with a message explaining why
+they were kicked. If enabled, a log message is also sent to the console. More exploit patches may be added in the
+future.
+
+Exploits Patched by Sponge Implementations
+==========================================
+
+#. Sign command exploit where a client could run a command such as 'op'
+#. Client could force the server to make the user respawn invisible
+#. Client could set an itemstack's display name and cause it to exceed the character limit
+
+.. warning::
+    The invisibility exploit patch has been disabled in recent Sponge builds due to the detection method falsely
+    accusing users of performing the exploit.
+
+Log Message Control
+===================
+
+Log messages for the exploit patches can be individually controlled in the Sponge config file. Please read the
+:doc:`../getting-started/configuration/sponge-conf` page for more information.
+
+.. tip::
+    Log messages can also be controlled via a command, instead of directly editing the config file. For example, to
+    enable the sign command exploit logging, type ``sponge config -g logging.exploit-sign-command-updates true`` in
+    the console (You can also type the commands in-game if you are an op).
--- a/source/server/management/index.rst
+++ b/source/server/management/index.rst
@@ -13,3 +13,4 @@ Contents
     bans
     permissions
     plugins
+    exploit-patches


source/server/management/exploit-patches.rst
```diff
@@ -0,0 +1,28 @@
...
+
+Exploits Patched by Sponge
+==========================
+
+#. Sign command exploit where a client could run a command such as 'op'
+#. Client could force the server to make the user respawn invisible
+#. Client could set an itemstack's display name and cause it to exceed the character limit
+
+.. warning::
+    The invisibility exploit has been disabled in recent Sponge builds due to false triggering.
```
saladoc: so the exploit cannot be used? or the fix for the exploit has been disabled? Also please clarify the "false triggering"

Tzky: My assumtion is that the exploit patch had too many false positives (players getting called cheater by the patch for no reason), thus the patch was temporarily disabled,

Tzky:
I suggest explaining where (in the config) you have to enable/disable the exploit patches. A link to the server-properties page might also suffice

Flibio:
@Tzky Maybe a link to the global.conf page instead?

Tzky:
Besides the minor typo, this looks good.

Inscrutable:
One problem I can see: It uses *Sponge* as shorthand for the *implementations*. This needs to be understood (or corrected). The patches should apply to both SpongeForge and SpongeVanilla, but Sponge usually refers to the entire project *per-se*.

Flibio:
@Inscrutable I updated the PR based on your comment. Let me know what you think.

Tzky:
Another suggestion: Please don't squash directly, instead wait until the PR is fully done. It's easier for us to spot the changes this way :)

source/server/management/exploit-patches.rst
```diff
@@ -0,0 +1,30 @@
...
+
+In recent Sponge builds (SpongeForge 974+), SpongeForge and SpongeVanilla patch a few client-server exploits. Whenever
+the implementations detect a user performing an exploit, they are kicked from the server with a message explaining why
+they were kicked. If enabled, a log message is also sent to the console. More exploit patches may be added in the
+future.
+
+Exploits Patched by Sponge Implementations
+==========================================
+
+#. Sign command exploit where a client could run a command such as 'op'
```
kenzierocks: Might want to include the conf entry here (and below), like `logging.exploit-sign-command-updates`, for quick and easy reference.

Flibio:
@kenzierocks I was planning on that, but instead left a link to the global.conf page where it was explained in more detail.

kenzierocks:
Ok, I guess that works. To me, if I was looking to disable it I would prefer if the keys were available immediately instead of having to scan the `global.conf` page.

Tzky:
I think showing the config options off on this page as well as linking to the global.conf page is fine.

Tzky:
@progwml6 suggested that we add a short notice *where* users should report exploits if they know of any. This will help us track them down and providing a fix/logs by default.

Suggested places to report:
* PM on forums
* PM on IRC
* better **not** post publicly on the forums if it's a still unknown issue

progwml6:
i would not reccomend IRC pm's as those can easily get lost/ignored when staff are busy.
an email adress that forwards to our developers might be a good option for this

Inscrutable:
This is probably a good place to include such a note, so that searching for 'exploit' takes you to this page. I'd be a little stronger with the wording too - "DO NOT post public messages anywhere detailing the exploit you found, that can make things worse. Don't even create a GitHub issue, notify us by PM or email, and keep it confidential until it's patched"

Tzky:
closed via 45fdf94 and  aa55eda

PR #419, opened 2016-01-03
Add Service Guidelines

### LINKED ISSUE #417: Document initialization process for services
@gravityfox suggested that we standardize the initialization process for services in general and especially for the EconomyAPI. This includes the init, pre-init and post-init events.

Flibio:
Adds service guidelines to the services page. Closes #417.

Let me know if the if the information I used is inaccurate.

### DIFF
--- a/source/plugin/services.rst
+++ b/source/plugin/services.rst
@@ -11,8 +11,21 @@ service manager:
 
 If you need to get an object reference to something, just get it off the service manager.
 
+Service Guidelines
+==================
+
+* Services should be registered during the ``POST_INITIALIZATION`` game state at the latest.
+* Services should be fully operational by the end of the ``SERVER_ABOUT_TO_START`` game state.
+
+You can read more about game states on the :doc:`lifecycle` page.
+
+.. note::
+
+    It is a good practice to register services as soon as possible so that other plugins can note that the service will
+    be provided.
+
 Providing your own service
---------------------------
+==========================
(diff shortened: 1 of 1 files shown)


Tzky:
You might want to link to https://docs.spongepowered.org/en/plugin/lifecycle.html

Flibio:
@Tzky Done.

saladoc:
@gabizou should probably take a look at this since we're basically describing a contract here.

Important thing: During the `GameInitializationEvent` plugins that want to use the service may determine if it is / will be provided and may store a reference if they want to keep one.

Flibio:
@saladoc Are we waiting for response from gabizou?

saladoc:
No we're not. I've seen a statement on when services should be ready, but I can't seem to find it. I do remember however, that being ready during the Initialization Phase was something that cannot be achieved.
After rereading the lifecycle docs as well I think Services should be registered by the `GamePostInitializationEvent` (Inter-plugin Communication should be ready to occur, Plugins providing an API accept baic requests). And fully operational by `GameAboutToStartServerEvent`

source/plugin/services.rst
```diff
@@ -11,8 +11,21 @@ service manager:
 
 If you need to get an object reference to something, just get it off the service manager.
 
+Service Guidelines
+==================
+
+* Services should be registered during the ``POST_INITIALIZATION`` game state.
```
saladoc: at the latest

PR #425, opened 2016-01-08
Added commonly requested timer example from the forums.

ryantheleach:
Many people have been enquiring on how best to cancel a timer from within a task/runnable. Hopefully this example will help ease that slightly.

### DIFF
--- a/source/plugin/scheduler.rst
+++ b/source/plugin/scheduler.rst
@@ -15,8 +15,9 @@ First, obtain an instance of the ``Scheduler``, and retrieve the ``Task.Builder`
 
     import org.spongepowered.api.scheduler.Scheduler;
     import org.spongepowered.api.scheduler.Task;
+    import org.spongepowered.api.Sponge;
 
-    Scheduler scheduler = game.getScheduler();
+    Scheduler scheduler = Sponge.getScheduler();
     Task.Builder taskBuilder = scheduler.createTaskBuilder();
 
 The only required property is the `Runnable <http://docs.oracle.com/javase/7/docs/api/java/lang/Runnable.html>`_,
@@ -29,6 +30,26 @@ which you can specify using ``Task.Builder#execute(Runnable runnable)``:
             logger.info("Yay! Schedulers!");
         }
     });
+
+or using Java 8 syntax with ``Task.Builder#execute(Runnable runnable)``
+
+.. code-block:: java
+
+    taskBuilder.execute(
+        () -> {
+            logger.info("Yay! Schedulers!");
+        }
+    );
+
+or using Java 8 syntax with ``Task.Builder#execute(Consumer<Task> task)``
+
+.. code-block:: java
+
+    taskBuilder.execute(
+        task -> {
+            logger.info("Yay! Schedulers! :" + task.getName());
+        }
+    );
 
 .. _task-properties:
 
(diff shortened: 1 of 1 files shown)


source/plugin/scheduler.rst
```diff
@@ -93,6 +93,35 @@ To cancel a task, simply call the ``Task#cancel`` method:
 
 	task.cancel();
 
+If you need to cancel the task, from within the runnable itself, you can instead opt to use a `Consumer<task>` in order to access the task from within.
+
+.. code-block:: java
+
+			Scheduler scheduler = game.getScheduler();
```
simon816: Now we can just use `Sponge.getScheduler()`, removes the assumption that the `game` variable exists.

source/plugin/scheduler.rst
```diff
@@ -93,6 +93,35 @@ To cancel a task, simply call the ``Task#cancel`` method:
...
+		}
+	}
+
+
+	private class CancellingTimerTask implements Consumer<Task> {
+		private int seconds = 60;
+		@Override
+		public void accept(Task task) {
+			seconds--;
+			Sponge  .getGame()
```
simon816: Some weird formatting here

kashike: Weird formatting throughout the example.

source/plugin/scheduler.rst
```diff
@@ -93,6 +94,35 @@ To cancel a task, simply call the ``Task#cancel`` method:
 
 	task.cancel();
 
+If you need to cancel the task, from within the runnable itself, you can instead opt to use a `Consumer<task>` in order to access the task from within.
```
12AwsomeMan34: \`Consumer\<task\>\` --> \`\`Consumer\<Task\>\`\`

Also, "If you need to cancel the task, from within the runnable itself, ..." sounds a bit clunky.
Maybe remove the first comma?
"If you need to cancel the task from within the runnable itself, ..."

12AwsomeMan34:
Tabs -> 4 spaces throughout the example

Tzky:
Some things require 2 spaces instead of 4 to work properly in sphinx, though.

RobertHerhold:
Looks good. Please update according to 12AwsomeMan34's suggestions & squash, and I'll merge

source/plugin/scheduler.rst
RobertHerhold: Maybe add a brief description of the below example: "The below example will schedule a task that will count down from 60 and cancel itself upon reaching 0" or something similar

ryantheleach:
I'll squash this as soon as the all clear is given on the Java 8 syntax that I introduced earlier in the article.

source/plugin/scheduler.rst
```diff
@@ -30,6 +31,26 @@ which you can specify using ``Task.Builder#execute(Runnable runnable)``:
...
+
+.. code-block:: java
+
+    taskBuilder.execute(
+        () -> {
+            logger.info("Yay! Schedulers!");
+        }
+    );
+
+or using Java 8 syntax with ``Task.Builder#execute(Consumer<Task> task)``
```
saladoc: Move this with the `Consumer<Task>` example. Also please change the example so that both consumer examples do the same thing.

source/plugin/scheduler.rst
```diff
@@ -93,6 +114,34 @@ To cancel a task, simply call the ``Task#cancel`` method:
...
+
+.. code-block:: java
+
+        Scheduler scheduler = Sponge.getScheduler();
+        Task.Builder taskBuilder = scheduler.createTaskBuilder();
+
+        Task task = taskBuilder.execute(new CancellingTimerTask())
+            .interval(1, TimeUnit.SECONDS)
+            .name("Self-Cancelling Timer Task").submit(plugin);
+    }
```
saladoc: this brace seems superfluous, since it has no matching opening brace. either wrap the above in a function or remove the brace

source/plugin/scheduler.rst
ryantheleach: Lamba's can't modify variables defined outside the lambda, so it's impossible as far as I know, to create an example that uses the timer. It's for this reason why I even bothered to define it as another class, otherwise I would have stuck to the java 8 syntax.

source/plugin/scheduler.rst
ryantheleach: That brace is the ending of the method that was initializing the class, before the inner class was defined. I wasn't sure to include the brace or not, or the method definition or not.

saladoc: either both or none.

source/plugin/scheduler.rst
```diff
@@ -93,6 +114,34 @@ To cancel a task, simply call the ``Task#cancel`` method:
 
 	task.cancel();
 
+If you need to cancel the task from within the runnable itself, you can instead opt to use a `Consumer<Task>` in order to
+access the task. The below example will schedule a task that will count down from 60 and cancel itself upon reaching 0.
+
+.. code-block:: java
+    @Listener
```
saladoc: I think you need a newline before `@Listener`

PR #433, opened 2016-01-12
Text (de)serialization documentation

### LINKED ISSUE #293: Document TextRepresentation
Include to/from
 * json
 * xml
 * legacy

RobertHerhold:
Resolves #293 

@zml2008 Can you look over the XML part of it and make sure it looks good?

### DIFF
--- a/source/plugin/text/representations/formatting-code-legacy.rst
+++ b/source/plugin/text/representations/formatting-code-legacy.rst
@@ -1,0 +1,62 @@
+===============================
+Formatting Code & Legacy Format
+===============================
+
+Text can be represented by using a special character indicating a formatting *of some kind* followed by a unique character
+indicating the specific formatting to be used. In the SpongeAPI, there are two different characters that are supported
+by default: the ampersand (&) and the section character (§).
+
+Both formatting schemes use the codes found at the `Minecraft Wiki <http://minecraft.gamepedia.com/Formatting_codes>`_.
+It is important to note that the Minecraft Wiki only displays the use of the section character (§), but the same codes
+work with the ampersand (&) as well.
+
+.. warning::
+    Text serialization using any kind of formatting code is limited to representing only a limited portion of the
+    capabilities of a ``Text``, namely formatting. It cannot represent any kind of click or hover actions. If brevity
+    is not an issue (as it is in Minecraft chat), it is recommended to use either the :doc:`./xml` or the :doc:`./json`.
+
+Ampersand Formatting
+====================
+By default, Sponge supports the formatting character `&` (ampersand). Using the ampersand format allows for easier user
+input of text formatting and is useful in such cases where brevity is necessary, such as in the Minecraft chat console.
+
+To use this formatting, you can access its corresponding ``TextSerializer`` with ``TextSerializers.FORMATTING_CODE``.
+From there, you can use the ``serialize`` or ``deserialize`` methods as normal:
+
+.. code-block:: java
+
+    import org.spongepowered.api.text.Text;
+    import org.spongepowered.api.text.format.TextColors;
+    import org.spongepowered.api.text.format.TextStyles;
+    import org.spongepowered.api.text.serializer.TextSerializers;
+
+    Text text = Text.of(TextColors.RED, TextStyles.UNDERLINE, "Hello World!");
+    String ampersandFormattedText = TextSerializers.FORMATTING_CODE.serialize(text);
+
+For example, the text "Hello World!", formatted with the color red and an underline would have the following representation
+with the ampersand formatting code: `&c&nHello World!`
+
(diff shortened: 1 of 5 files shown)


Spongy:

saladoc:
Mention *why* legacy format is discouraged (afaik because it has no support for text actions).

source/plugin/text/representations/xml.rst
```diff
@@ -0,0 +1,51 @@
+===============
+TextXML Format
+===============
+
+Sponge uses a proprietary form of `XML <https://en.wikipedia.org/wiki/XML>`_ called "TextXML" to denote the coloring
```
ryantheleach: proprietary is a bit strong of a word, considering it wasn't sponges to begin with but adapted with permission from @TomyLobo

The differences that I am aware of  between TextXML and XML is the schema is already defined, you don't need a wrapping root node for a valid fragment(Sponge adds this automatically), and whitespace is counted.

RobertHerhold:
@saladoc the FORMATTING_CODE serializer (not the legacy one) also wouldn't then...

TomyLobo:
I'm gonna cause a paradox by restricting anyone using it from calling it proprietary :D

TomyLobo:
What's missing is the onClick and onHover attributes (or whatever you renamed them to)

RobertHerhold:
@TomyLobo Thanks for that, fixed now.

saladoc:
I'm still not happy with the Legacy format. It should be more clear that 
1) both the ampersand and the legacy format (which are essentially the same tbh) are deprecated
2) those formats are deprecated because they are only able to represent a limited portion of the capabilities of a `Text`, namely only formatting

source/plugin/text/representations/formatting-code-legacy.rst
```diff
@@ -0,0 +1,46 @@
...
+
+For example, the text "Hello World!", formatted with the color red and an underline would have the following representation
+with the ampersand formatting code: `&c&nHello World!`
+
+Legacy Formatting
+=================
+
+The legacy text representation is the format widely used in older versions of Minecraft, represented by the section
+character (§). Sponge provides serialization and deserialization using the legacy format **only for compatibility**.
+It should not be used unless absolutely necessary. Rather, it is recommended to use formatting with the ampersand, as
```
saladoc: why is ampersand recommended? It should only be used to allow a user to enter a formatted text / to parse legacy configs. for storing text a proper format should be used

RobertHerhold:
I'd definitely willing to revise it, but the ampersand formatting isn't deprecated in the API, so I figured there would be some other reason legacy was deprecated

saladoc:
oh well If it is not deprecated in the API, then it's not deprecated. Then you only need to make sure to include its limitations.

ryantheleach:
Well the only real difference between ampersand and section symbol is all the potenital utf8 issues that the section symbol can cause in configs, and the difficulty of users entering it into the configs.

TomyLobo:
I just press shift+3 :P

Minecrell:
Formatting codes using `§` were the original one that are now being more and more removed from Minecraft in favor of the new JSON chat format. The primary point why it is completely deprecated is: While it is still necessary (especially internally), sending color codes of this kind directly (without pulling it through the legacy text serializer) is still "kind of" working, but these colors will for example break after a line break.

The client will never display `§` normally, so unlike when using `&` you will always get weird output when the text serializer is not used properly. Additionally the ampersand color codes have never been a real format used by Minecraft, so it is fine to provide them as an additional, independent text serialization format just like TextXML exists in SpongeAPI without deprecating it (even though the limitations it has must be considered when using it).

I can still see still major uses for the ampersand formatting codes, for example if you want to colorize chat messages (sent by a player in chat) where a character limit exists. Additionally the formatting codes is what people are used to for chat messages so it wouldn't make sense preventing them from using them.

ryantheleach:
I'm confused at why pulling it through the legacy text serializer is deprecated then, if it's using it everywhere else in sponge that is the problem?

Minecrell:
@ryantheleach Basically, if you'd need to decide to make one of them recommended, the ampersand formatting codes are just the better option because they never cause weird issues with charsets (or the input environment, e.g. you can't type `§` in chat), or problems when they're not properly pulled through the legacy text serializer.
The deprecation here is not really to indicate this will be likely removed in the future, but just that we recommend plugin authors to use the ampersand formatting codes if you decide between those two, because they provide better usability.

TomyLobo:
I agree with deprecating it to discourage active use.
However, make it clear in the deprecation comment that it won't be removed unless there's a really good reason.

RobertHerhold:
Updated according to feedback. Any other comments/suggestions before I merge?

PR #441, opened 2016-01-21
Update global.conf

12AwsomeMan34:
Note that this pr targets the 3.1.0 version of the docs, for when the docs update to 3.1.0.

I will check through it and update where necessary when 3.1.0 is officially released, for now consider this a start.


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/68568eec/index.html.

- modified https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/68568eec/server/getting-started/configuration/sponge-conf.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/441#issuecomment-173423825), not to the direct url to the preview._

Inscrutable:
Two things I think need some attention here:
Aliases - it's now a nested config, and the example is out of date. There's a better walk through [here](https://forums.spongepowered.org/t/conflicting-plugin-commands-look-no-further/11011) although I'm not sure it's 100% accurate. Is it meant to be a : or = as separator? Also, the existing example has a space after the :
I also note that the entities list in the options seems to have been shrinking... I think I originally had all vanilla entities, but they're gone. That list self-populates as you encounter them, so it isn't complete at the start anyway. It might help if we offer a complete list (or link to it).

12AwsomeMan34:
1. I took that aliases thing from the old aliases section on that page. I've tested, both seem to work, however when the server starts it changes : to an = (not when directly reloading the config). So it's not a huge issue at present, could likely be changed in the next commit to this pr.
2. I take it from a freshly generated config as it implies at the top of the page, although that could be changed to something along the lines of "The below is a complete config with all possible nodes. It was built on SpongeForge build ... Note that certain sections will not be present immediately, and will be added to the server config when the server encounters them.". I'll dig through the history and find the entities and what not that was originally present.

PR #443, opened 2016-01-26
Document server timings

### LINKED ISSUE #380: Document Aikars timings
Aikar's Timings Viewer (https://github.com/aikar/timings) got included into Sponge, so we need to document:
 - what it is
 - what it does
 - usage examples
 - configuration (if available?)
 - limitations (what isn't tracked)
 - how plugins can implement their own timings (thanks @ryantheleach )

12AwsomeMan34:
Note that this is only the server use of timings, not for plugins, as such it only partially resolves #380.

Point out if anything was missed/incorrect/needs expanding/needs better explanation.

### DIFF
--- a/source/server/spongineer/commands.rst
+++ b/source/server/spongineer/commands.rst
@@ -43,9 +43,7 @@ Command               Description                               Permission
                       config.
 /sponge version       Prints the Sponge/SpongeAPI versions to   sponge.command.version
                       the console.
-/sponge timings       The command ``/sponge timings report`` 
-                      generates a report on your server          sponge.command.timings
-                      performance at http://timings.aikar.co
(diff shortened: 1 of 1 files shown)


Spongy:

source/server/spongineer/commands.rst
```diff
@@ -81,6 +79,35 @@ Command               Description                               Permission
 
     This would alter the config of world named DIM1.
 
+Timings
+~~~~~~~
+
+Timings are a tool built into Sponge that allows server administrators to monitor the performance of their server.
+Timings will collect information about a server so that a report may later be generated on the data. Things that are
```
RobertHerhold: I'd try to avoid "Things". Maybe "Information" or "Data"

PR #445, opened 2016-02-01
Better item docs

12AwsomeMan34:
Expands on the previously basic item docs. Now includes creating an `ItemStack`, spawning an `ItemStack` through an `Item` entity, and creating an `ItemStack` from a `BlockState`. If there is anything else that should be added, point such out.

### DIFF
--- a/source/plugin/items/creating.rst
+++ b/source/plugin/items/creating.rst
@@ -1,0 +1,50 @@
+=====================
+Creating an ItemStack
+=====================
+
+Let's create a Diamond Sword:
+
+.. code-block:: java
+
+    import com.google.inject.Inject;
+    import org.spongepowered.api.Game;
+    import org.spongepowered.api.data.key.Keys;
+    import org.spongepowered.api.data.manipulator.mutable.item.EnchantmentData;
+    import org.spongepowered.api.data.meta.ItemEnchantment;
+    import org.spongepowered.api.item.Enchantment;
+    import org.spongepowered.api.item.ItemTypes;
+    import org.spongepowered.api.item.inventory.ItemStack;
+    import org.spongepowered.api.text.Texts;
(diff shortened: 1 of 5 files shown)


Spongy:

source/plugin/items/creating.rst
```diff
@@ -0,0 +1,120 @@
...
+    import java.util.Optional;
+    
+    public void spawnItem(ItemStack superMegaAwesomeSword, Location<World> spawnLocation) {
+        Extent extent = spawnLocation.getExtent();
+        Optional<Entity> optional = extent
+            .createEntity(EntityTypes.ITEM, spawnLocation.getPosition());
+        if (optional.isPresent()) {
+            Entity item = optional.get();
+            item.offer(Keys.REPRESENTED_ITEM, superMegaAwesomeSword.createSnapshot());
+            extent.spawnEntity(item, Cause.of(EntitySpawnCause.builder()
```
gabizou: Note that this also requires that it's a `NamedCause.source(SpawnCause)`

Inscrutable:
It looks pretty good to me, nothing serious to quibble about. If the code examples are OK (and not API 3.10+), it's good to go. Gabi's note on 4.0 can wait.

12AwsomeMan34:
One of my questions is on the spawning bit. It seems users on the forums are pointing out that spawn causes are not really implemented in 3.0 (though the api for it is there). Should we just leave a note somewhere under the spawning examples clearly stating that as of 3.0, it is not implemented and thus they should simply specify the command source or something and tell them that they should update to use spawn causes when possible? Would probably save plugin devs some time.

Inscrutable:
That sounds fair. We are allowed to admit that some stuff isn't implemented - is it absent, or partial? Recommending the API version where it is complete would not be a bad idea, although we may not be able to include a link yet. =)

PR #446, opened 2016-02-06
IntelliJ + Gradle Requires Additional Steps

dkolb:
Without these steps, debugging fails because IntelliJ can't resolve any references to classes, annotations, etc. in the SpongAPI or other dependencies.


Tzky:
Thanks for reporting that one! :)

Issue #437, opened 2016-01-16
Ongoing Minor Issue List (OCD List) SpongeDocs Edition


Tzky:
I'd like to start an OCD post on the Docs repo, this way we don't push minor fixes but instead collect them and push them in bunches.

@SupaHam reported these two in '#sponge:
- [x] in `/plugin/debugging` phrasing in L91: 
  - [x] mention of Sponge(Vanilla) and missing SpongeForge
  - [x] "the sentences don't make any sense in this particular order" 
- [x] in `/plugin/event/causes`: `Optionml<Entity>` in L45

according to PR#447:
- [x] in `plugin/optional/usage.rst` (L166):  `if (optinalString.isPresent())`

I'll quote @phroa here:
- [x] the link to GenericArguments' source code on the argument parsing section is busted, should be https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/command/args/GenericArguments.java

@Dannyps reported:
- [x] in `server/getting-started/configuration/index` : `Saving a world config may be desired *after making modifications to the file*.` is confusing. It should read: `
  Saving a world config to the file may be desired after making modifications.`

Deamon5550:
in `/plugin/wgen/modifiers` the example for registering the world generator modifer needs to be updated to `Sponge.getRegistry().register(WorldGeneratorModifier.class, new MyModifier());`

Tzky:
I'll quote @phroa here:
- [ ] the link to GenericArguments' source code on the argument parsing section is busted, should be https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/command/args/GenericArguments.java

Tzky:
according to PR#447:

- [ ] in ``plugin/optional/usage.rst ( L#166):
   ``if (optinalString.isPresent()) {``

Tzky:
@Dannyps reported:
- [ ] in ``server/getting-started/configuration/index`` : ``Saving a world config may be desired *after making modifications to the file*.`` is confusing. It should read: ``
Saving a world config to the file may be desired after making modifications.``

12AwsomeMan34:
- [ ] The page [about/introduction](https://docs.spongepowered.org/en/about/introduction.html) mentions that the project leaders are blood, sk89q, and Zidane. This should be updated in accordance with [about/staff](https://docs.spongepowered.org/en/about/staff.html).

RedNesto:
[...]

PR #447, opened 2016-02-08
Update usage.rst

Pocketkid2:
Sorry I just had to


ZephireNZ:
This would be better suited in #447

ZephireNZ:
Whoops, that should be #437

Tzky:
Correct, we'll collect these typos and either:
a) merge them when the next version bump occurs or
b) when there's enough to waste a commit on OCD 

BUT: Thanks for reporting this one, any help is appreciated! :)

PR #452, opened 2016-02-23
Add Concepts section + Talk about traversable nodes.

20zinnm:
Moved Optionals to a section dedicated to concepts, and added a 'Data' section under Concepts dedicated to different data storage mechanisms and theory.

Sponge supports development best-practices and that alone makes this section appropriate; should Sponge wish to encourage best-practices among its community, it should encourage a section dedicated to design models in order to improve the general knowledge of the community.


Inscrutable:
It's nice to have some explanations and good practice examples around, but it really has to be targeted at Sponge to be relevant to the docs. Adding general java tutorials is out of scope. SpongeDocs isn't meant to be a java tutorial. We have a "best practices" page for the more critical stuff, but even that's mostly Sponge-specific.

20zinnm:
@Inscrutable That's why I moved it to a separate section. I see the point, but if Sponge is focused on performance then its plugins should be too. Maybe there should be a section of the forums dedicated to this stuff?

Inscrutable:
Yeah, somewhere in Plugin Development would work OK, I think there are a few other code-style guides in there too. If one of the dev staff feels it's appropriate for "best practices", we can put a short rant in there too. However, in general, java tutorials are a hard limit for SpongeDocs.

Tzky:
I have to agree with @Inscrutable. While it's nice to have everything on the Docs, there is a limit on what shouldn't be there. I appreciate your effort and the idea of this PR, but it'd be better to post this on the forums instead. 

If we need a ``guides/ tutorials`` section on the forums will depend on a) the content we can post there, b) the need for this section and of course c) if Sponge staff actually wants to introduce said section. However that's something that Docs staff can't decide ;)

PR #453, opened 2016-02-28
Touched up scheduler documentation.

minecrafter:
This is a small change that allows us to better explain asynchronous tasks.

AFAIK, the Sponge scheduler should be thread-safe.

PR #467, opened 2016-03-21
ExecutorService Documentation

Kiskae:
Decided to contribute documentation about the ExecutorService abstraction available through the Scheduler since I contributed that part of the code, currently a WIP but the goal is to show how it is used in a few popular concurrent frameworks.

Open to suggestions/comments on things that should be clarified

### DIFF
--- a/source/plugin/scheduler.rst
+++ b/source/plugin/scheduler.rst
@@ -143,6 +143,8 @@ access the task. The below example will schedule a task that will count down fro
         }
     }
 
+.. _asynchronous-tasks:
+    
 Asynchronous Tasks
 ~~~~~~~~~~~~~~~~~~
 
@@ -167,6 +169,8 @@ In addition, there are a few other operations that are safe to do asynchronously
 Compatibility with other libraries
 ==================================
 
+:ref:`asynchronous-tasks`
+
 TODO: larger plugins mean you probably want to do some work asynchronously, note about standard java concurrency interfaces and thread-safety
 
 TODO: interfaces for both sync and async scheduling, sync is the most interesting.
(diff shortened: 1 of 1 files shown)


source/plugin/scheduler.rst
```diff
@@ -163,3 +163,80 @@ In addition, there are a few other operations that are safe to do asynchronously
...
+    SpongeExecutorService executor = Sponge.getScheduler().createSyncExecutor(plugin);
+
+    CompletableFuture.supplyAsync(() -> {
+        // *Async methods default to running on ForkJoinPool.commonPool()
+        return /* awesome value */
+    }).thenAcceptAsync((awesomeValue) -> {
+        /* use awesomeValue on the main thread */
+    }, executor); // Run this future on our executor
+
+RxJava
```
kashike: A link to this should be provided.

source/plugin/scheduler.rst
```diff
@@ -163,3 +163,80 @@ In addition, there are a few other operations that are safe to do asynchronously
...
+
+.. code-block:: java
+
+    import rx.Scheduler;
+    import rx.schedulers.Schedulers;
+
+    SpongeExecutorService executor = Sponge.getScheduler().createSyncExecutor(plugin);
+    Scheduler minecraftScheduler = Schedulers.from(executor);
+
+    Observable<T> someObservable = createObservable();
```
kashike: Where is the `createObservable` method coming from?

Kiskae: @kashike User-defined, might just replace it with `<define your observable>` or something similar

source/plugin/scheduler.rst
Kiskae: @kashike I dont think there is an example of external project integration in the documentation that I can base this section on, I presume adding a link to the title would break the documentation? :P

Kiskae:
First version of the documentation is now ready for review :)

source/plugin/scheduler.rst
```diff
@@ -143,6 +143,8 @@ access the task. The below example will schedule a task that will count down fro
         }
     }
 
+.. _asynchronous-tasks:
```
12AwsomeMan34: https://github.com/SpongePowered/SpongeDocs/pull/453#discussion_r54659546

source/plugin/scheduler.rst
```diff
@@ -163,3 +165,139 @@ In addition, there are a few other operations that are safe to do asynchronously
...
+==================================
+
+.. _ExecutorService: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html
+
+As your plugin grows in size and scope you might want to start using one of the many concurrency libraries available 
+for Java and the JVM.
+These libraries do tend to support Java's ExecutorService_ as a means of directing on which thread the task is executed.
+
+For compatibility Sponge can create an ``ExecutorService`` using the ``Scheduler#createSyncExecutor(Object plugin)`` or 
+``Scheduler#createAsyncExecutor(Object plugin)`` methods which executes tasks on Sponge's synchronous or asynchronous
```
12AwsomeMan34: Bit of a run-on sentence.
For compatibility, Sponge can create an \`\`ExecutorService\`\` using the ... or ... methods, which executes tasks on Sponge's ...

source/plugin/scheduler.rst
```diff
@@ -163,3 +165,139 @@ In addition, there are a few other operations that are safe to do asynchronously
...
+Almost all libraries have some way of adapting the ``ExecutorService`` to natively schedule tasks.
+As an example the following paragraphs will explain how the ``ExecutorService`` is used in a number of libraries.
+
+CompletableFuture (Java 8)
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. _CompletableFuture: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html
+
+With Java 8 the CompletableFuture_ object was added to the standard library.
+Compared to the ``Future`` object this allows for the developer to provide a callback that is called when the future
```
12AwsomeMan34: Minor nitpick,
Compared to the \`\`Future\`\` object, this allows for ...

source/plugin/scheduler.rst
```diff
@@ -163,3 +165,139 @@ In addition, there are a few other operations that are safe to do asynchronously
...
+completes rather than blocking the thread until the future eventually completes.
+
+CompletableFuture_ is a fluent interface which usually has the following 3 variations for each of its functions:
```
12AwsomeMan34: 3 should probably be spelled out.
'three'

source/plugin/scheduler.rst
```diff
@@ -163,3 +165,139 @@ In addition, there are a few other operations that are safe to do asynchronously
 
 * Independent network requests
 * Filesystem I/O (excluding files used by Sponge)
+
+Compatibility with other libraries
+==================================
+
+.. _ExecutorService: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html
```
12AwsomeMan34: From the looks of it, `ExecutorService_` is only ever referenced once. Wouldn't it just be better to do
`ExecutorService <https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html>`
in the text itself?

Kiskae:
@12AwsomeMan34 Thanks for the feedback :smile:

Inscrutable:
There are quite a few comments inside code blocks, too. Anything inside a code block does not get translated, so if we can minimise that it would help with translations. We realise that some are inevitable.

Kiskae:
@Inscrutable I see what you're talking about, but besides some advice about implicits in the scala section I don't see how I could move those comments out of their context, since most of them are about where the specific piece of code is being executed.

Kiskae:
I believe I've made it so the comments contained in the code fragments illustrate their point regardless of translations (by purely making them about SYNC/ASYNC work)

Inscrutable:
If this is OK by you @12AwsomeMan34 then I think it's ready to merge.

source/plugin/scheduler.rst
```diff
@@ -163,3 +163,149 @@ In addition, there are a few other operations that are safe to do asynchronously
...
+
+.. code-block:: java
+
+    import java.util.concurrent.CompletableFuture;
+    
+    SpongeExecutorService minecraftExecutor = Sponge.getScheduler().createSyncExecutor(plugin);
+
+    CompletableFuture.supplyAsync(() -> {
+        // ASYNC: ForkJoinPool.commonPool()
+        return 42
```
12AwsomeMan34: No semicolon in this?

Kiskae: @12AwsomeMan34 Thats what I get for working in 4 different JVM languages at once...

PR #484, opened 2016-05-02
It is now recommended to use a qualified plugin ID

Advancid:


jamierocks:
Nope - not anymore ;)

Minecrell:
This will be changed back soon, so the docs were updated to our new recommendation directly.

PR #491, opened 2016-05-13
Make titles more concise

RobertHerhold:
This PR makes many of the articles' titles shorter. I believe that it should be a primary concern when writing documentation to write it in as few words as possible, and also it should be immediately obvious what article you are looking for, rather than reading "Working with..." every single time.

Input is appreciated


Spongy:

PR #510, opened 2016-06-13
Update custom wgen docs for populator area change

Deamon5550:
This is a small update to clarify the use of the new parameter passed to populators in spongeapi bleeding right now (Ref Spongepowered/SpongeAPI#1231)


Spongy:

PR #513, opened 2016-07-02
Plugins shouldn't have Sponge in their name

Inscrutable:
We might want to include this as a note in the plugins section too.

### DIFF
--- a/source/about/rules.rst
+++ b/source/about/rules.rst
@@ -51,6 +51,9 @@ list of do's and don'ts. This stuff should all be common sense.
   right to remove the plugin listing. What constitutes as a for-sale "lite" version will be decided at the discretion
   of Sponge Staff.
 
+* Please don't include *Sponge* in your plugin name. We reserve the use of *Sponge* for official SpongePowered works.
+  There is one exception: you may append "for Sponge" to the title if your plugin also has versions for other APIs. 
+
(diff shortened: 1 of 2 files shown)


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/e65aaea7/index.html.

- modified https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/e65aaea7/about/rules.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/513#issuecomment-230081343), not to the direct url to the preview._

RobertHerhold:
It should probably be enforced through Ore as well, rejecting plugins with "Sponge" in the name

zml2008:
That's hard to do correctly -- what if the plugin actually deals with sponges the block in MC? Then a name containing Sponge could be totally appropriate.

Inscrutable:
Indeed. The main reason I put this as a PR is to harvest opinions on what constitute exceptions. We generally don't want a plethora of third-party plugins with Sponge in the name, as it will confuse all sorts of folks. So far the exceptions seem to be:
* Ported Plugins with versions for other APIs
* Plugins dealing specifically with the minecraft block "Sponge". These will probably be few.

Are there any other reasonable exceptions anyone can think of?

Inscrutable:
Suggested pages for extra mentions of this rule: _Plugin Identifiers_ and/or _Best Practices_.

PR #529, opened 2016-09-06
Fix code example

jamierocks:


Tzky:
Is this a 5.x fix or is this also broken in 4.x?

jamierocks:
The example referenced a parameter `dataHolder`, that didn't exist the intended parameter was `target`. Perhaps my commit message was better off as Fix Broken code example.

PR #531, opened 2016-09-11
Change idea.rst to show steps...

Redrield:
...For using DemonWav's IntelliJ plugin to make a project.

### DIFF
--- a/source/plugin/workspace/idea.rst
+++ b/source/plugin/workspace/idea.rst
@@ -12,8 +12,12 @@ Creating your project
 ~~~~~~~~~~~~~~~~~~~~~
 
 * Open **IntelliJ IDEA**.
+* Click ``Configure``, then click ``plugins``.
+* Click ``Browse Repositories``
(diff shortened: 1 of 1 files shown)


kinggoesgaming:
I think think this should be added as a separate page. Simply because not everyone might want to use minecrqft development plugin

Redrield:
Fair enough. I'll close this for now and open up a new one on a separate page

RobertHerhold:
I don't believe this belongs in the Sponge documentation since this IntelliJ plugin is not endorsed or developed by Sponge. @Inscrutable please confirm

Meronat:
Well I mean some Sponge developers have been working with the lead developer and the plugin is fairly useful. Once it gets out of alpha, maybe it would be a good choice to endorse it, as it does simplify the whole process a little and provides some small but nice features.

I don't think it should necessarily be the main or only way we should teach people how to set up their projects, but a nod towards it and a little guide wouldn't harm anyone.

Issue #32, opened 2014-12-23
Permissions API Page for Plugins


Inscrutable:
We need a walk-through of the use of the Permissions API on the page source/plugins/permissions. Best PR gets an imaginary cookie.

gabizou:
Well, [Permissions](https://github.com/SpongePowered/SpongeAPI/pull/340) hasn't been accepted yet, so whoever is writing documentation would probably need to wait to do so.

boformer:
API is now ready. I will work on this next weekend.

ryantheleach:
Made any progress?

RobertHerhold:
@ryantheleach Go ahead. I don't think anybody is working on it currently.

ryantheleach:
https://github.com/SpongePowered/SpongeAPI/commit/81bba65212a29885a5306ef6d286316213d62ed7

Permission Descriptions have made it into the API.

ryantheleach:
related: https://github.com/SpongePowered/SpongeDocs/issues/136

ryantheleach:
Some musings and definitions that I wrote while thinking about what needs to be covered to understand permissions in their entirety.

What is a permission?

A permission, is a right to do some action. They are represented as a permission node, together with a tristate value.

A permission "node" is a permission id. they are in the form of exampleplugin.command.use 

Subjects are things that have rights / permissions assigned to them, and have an ever changing context (they are contextual) usually these are users or players, but could be templates or groups.

A context can be of any number of things, A world a player is in, the current dimension, the hostname they are connected to, the region a player is in, their gamemode, anything that provides a contextCalculator can be a context. Contexts can be used by a PermissionsService to aid decision making in whether a subject has a given permission.

A permission check, is a test to see if a subject, has the right to perform an action at the present time, in a given context.

A tristate value, is a value of true, false or undefined.

A tristate value of true grants the subject a permission, false will deny it. Undefined varies, but usually implies that a fallback of some sort will happen, according to rules in the implementation of the permissions service.

Is setting a permission to undefined it the same as clearing a permission?
What is a parent?

A SubjectCollection is simply a collection of subjects, examples of which is the collection of all Users, a Collection of groups, a subset of groups and or users.

A PermissionsDescription is a hint to the PermissionsService on who should have what permission by default, it also contains a Text description which aids end users, and where possible should contain a ClickAction link to documentation.

A NodeTree is a tree representation of a stored hierarchy of permissions.

[...]

Issue #136, opened 2015-03-10
Mention default data of permission service


boformer:
This is quite important for Bukkit plugin ports, and also useful for other plugins.
- Create permissions that are true by default
- Disable implicit inheritance

https://github.com/SpongePowered/SpongeAPI/issues/494#issuecomment-78165659

BlackGoat:
Thus far, Issue #32 (writing about the Permissions API) hasn't had any contributions. The stub page is at source/plugins/permissions.rst for reference. This explanation would belong there too.

Inscrutable:
Bumping this because it could really use some attention.
@zml2008 will you be able to help us get this out of the way?

zml2008:
Not right away -- also there are a few changes to the permissions API coming in in v5 (regarding default data specifically) that would be useful to incorporate.

ST-DDT:
Is this covered by: https://github.com/SpongePowered/SpongeDocs/pull/536
Or is there anything missing for that?

Issue #434, opened 2016-01-14
Itemstack Trade Offer API


Inscrutable:
A new feature for handling Itemstacks & trade offers. Coming soon, and will need docs.
https://github.com/SpongePowered/SpongeAPI/pull/1032/files#r49403140

12AwsomeMan34:
This was recently moved from 4.0 to 5.0.

ST-DDT:
Just a little link to the API stuff: https://github.com/SpongePowered/SpongeAPI/tree/bleeding/src/main/java/org/spongepowered/api/item/merchant

PR #536, opened 2016-10-10
Permission documentation

ST-DDT:
# Permission System

Creates some documentation on the permission system. Fixes: https://github.com/SpongePowered/SpongeDocs/issues/32 , https://github.com/SpongePowered/SpongeDocs/issues/136

If you would like to help me just go ahead.

## TODO
- [x] Definitions
- [x] Check for permissions
- [x] Provide descriptions
- [x] Provide contexts

## TBD
- [x] Specify precedence for persistent and transient ``SubjectData``
  - [Inheritance-Note](https://github.com/SpongePowered/SpongeDocs/pull/536/files#diff-47577673bb5da462c61537e29ed1ed0aR171)
  - [Inheritance-Self](https://github.com/SpongePowered/SpongeDocs/pull/536/files#diff-47577673bb5da462c61537e29ed1ed0aR181)
  - [Inheritance-Parent](https://github.com/SpongePowered/SpongeDocs/pull/536/files#diff-47577673bb5da462c61537e29ed1ed0aR182)
  - [SubjectData-Precedence](https://github.com/SpongePowered/SpongeDocs/pull/536/files#diff-47577673bb5da462c61537e29ed1ed0aR233)

## PS

Please note that this is my first participation on this docs, so I would appreciate early feedback and suggestions for improvements.

## Links
- [SpongeAPI-Permissions](https://github.com/SpongePowered/SpongeAPI/tree/bleeding/src/main/java/org/spongepowered/api/service/permission)
- [API-Issue: Documentation & Stricter Specifications](https://github.com/SpongePowered/SpongeAPI/issues/1383)

### DIFF
--- a/source/plugin/permissions.rst
+++ b/source/plugin/permissions.rst
@@ -2,9 +2,371 @@
 Permissions
 ===========
 
+.. javadoc-import::
+    org.spongepowered.api.command.CommandSource
+    org.spongepowered.api.entity.living.player.Player
+    org.spongepowered.api.service.context.Context
+    org.spongepowered.api.service.context.ContextCalculator
+    org.spongepowered.api.service.permission.PermissionDescription
+    org.spongepowered.api.service.permission.PermissionService
+    org.spongepowered.api.service.permission.Subject
+    org.spongepowered.api.service.permission.SubjectCollection
+    org.spongepowered.api.service.permission.SubjectData
+
+Permission
+==========
+
+A permission is a case-insensitive hierarchical string key that is used to determine whether a Subject can do a specific
+action or not.
+The string is split into separate parts using the dot character.
+Permissions should be structured like this.
+
+.. code-block:: text
+    
+    <PluginID>.<MainGroup>.<Subgroup1>...
+
+Allowed characters are:
+
+* "A" - "Z"
+* "a" - "z"
+* "0" - "9"
+* "_"
+* "-"
+* "."
+
+Inheritance
+~~~~~~~~~~~
+
+If a user has the permission ``myPlugin.commands`` then they automatically have all sub permissions such as
+``myPlugin.commands.teleport`` unless the permissions are explicitly removed.
 
 .. note::
 
-    This page has not been filled yet by our Documentation Team. If you feel like you can help, you can do so on `our
-    GitHub repository <https://github.com/spongepowered/spongedocs>`_. Also see the `related GitHub Issue
-    <https://github.com/SpongePowered/SpongeDocs/issues/32>`_ for more information on what is required.
+    There is no such thing as a ``myPlugin.commands.*`` wildcard permission.
+    Use ``myPlugin.commands`` for that.
+
+Structure-Example
+~~~~~~~~~~~~~~~~~
+
+The following example show one possible way of structuring permissions, but following this structure is not required at
+all.
+
+* ``myPlugin``
+    * Grants **full** access to all plugin's permissions.
+
+* ``myPlugin.commands``
+    * Grants **full** access to all plugin's commands.
+
+* ``myPlugin.commands.teleport.execute``
+    * Only grants the user the permission to execute the command. Without this permission he is not able to execute the
+      command even if he has other `teleport` permissions.
+      (With this permission alone the user would only be able to teleport himself to others in his current world.)
+
+* ``myPlugin.commands.teleport.all``
+    * Only grants the user the permission to teleport all players at once.
+
+* ``myPlugin.commands.teleport.worlds``
+    * Only grants the user the permission to teleport to all worlds.
+
+* ``myPlugin.commands.teleport.worlds.mynetherworld``
+    * Only grants the user the permission to teleport to *mynetherworld*.
+
+PermissionDescription
+=====================
+
+The :javadoc:`PermissionDescription` is an utility that is meant to provide the server owner with details on a certain
+permission. ``PermissionDescription``\s are an optional feature a :javadoc:`PermissionService` can provide. The creation
+of a PermissionDescription does **not** have any impact on whether a permission exists, who has access or what its
+default value is.
+
+The description consists of:
+
+* the target permission id
+* a Text description
+* one or more assigned roles
+* the owning plugin
+
+If you have a dynamic element such as a ``World`` or ``ItemType`` then you can use ``<TemplateParts>``.
+
+Usage-Example
+~~~~~~~~~~~~~
+
+.. code-block:: text
+
+    import org.spongepowered.api.service.permission.PermissionDescription;
+    import org.spongepowered.api.service.permission.PermissionDescription.Builder;
+    import org.spongepowered.api.service.permission.PermissionService;
+    import org.spongepowered.api.text.Text;
+    import java.util.Optional;
+
+    Optional<Builder> optBuilder = permissionService.newDescriptionBuilder(myPlugin);
+    if (optBuilder.isPresent()) {
+        Builder builder = optBuilder.get();
+        builder.id("myPlugin.commands.teleport.execute")
+               .description(Text.of("Allows the user to execute the teleport command."))
+               .assign(PermissionDescription.ROLE_STAFF, true)
+               .register();
+    }
+
+Simple-Result
+~~~~~~~~~~~~~
+
+.. code-block:: text
+    
+    myPlugin.commands.teleport.execute
+    
+    Description: Allows the user to execute the teleport command.
+    Role: user
+    Owner: MyPlugin v1.2.3
+
+Template-Result
+~~~~~~~~~~~~~~~
+
+.. code-block:: text
+    
+    myPlugin.commands.teleport.worlds.<World>
+    
+    Description: Allows the user to teleport to the world <World>.
+    Role: staff
+    Owner: MyPlugin v1.2.3
+
+.. tip::
+
+    You might skip writing descriptions for some parent permission groups such as ``myPlugin.commands.teleport.worlds``
+    or ``myPlugin.commands`` as their meaning can be derived from the permission structure and the defined children
+    alone.
+
+Subject
+=======
+
+A :javadoc:`Subject` is a holder of assigned permissions. It can be obtained from the ``PermissionService`` via 
+:javadoc:`SubjectCollection`\s.
+:javadoc:`CommandSource`\s such as :javadoc:`Player`\s are ``Subject``\s by default, but there are many other types of
+``Subject``\s. Anything that has permissions is a Subject even if it just delegates the checks to a contained Subject.
+Permissions can be granted or denied to a Subject. If a permission is neither granted nor denied its setting will be
+inherited. See Inheritance.
+Subjects provide methods to check whether they have a certain permission or not.
+Plugins that use this method should only query for the specific permission they want to check. It is the
+PermissionService's task to respect the permission and subject inheritance.
+
+Example
+~~~~~~~
+
+The following example could be used to check whether the Player is allowed to execute the teleport command.
+
+.. code-block:: java
+
+    import org.spongepowered.api.entity.living.player.Player;
+    import org.spongepowered.api.world.World;
+
+    public boolean canTeleport(Player subject, World targetWorld) {
+        return subject.hasPermission("myPlugin.command.teleport.execute") 
+                && (subject.getWorld() == targetWorld
+                        || subject.hasPermission("myPlugin.command.teleport." + targetWorld.getName()));
+    }
+
+Inheritance
+~~~~~~~~~~~
+
+If a ``Subject`` has a permission assigned, it will use that value.
+Otherwise it will be inherited from any parent ``Subject``. It does not matter what kind of parent (e.g. group or 
+player) ``Subject`` that might be.
+
+If neither the subject itself nor any parent subjects grant or deny a permission then it will be inherited from the
+default ``Subject``\s. Each ``SubjectCollection`` defines its own defaults. The global and weakest default subject can be
+obtained from the ``PermissionService``. Plugins may define their own permissions to the default's transient
+:javadoc:`SubjectData` during every server start-up. This allows server owners to overwrite the defaults defined by
+plugins according to their needs using the default's persistent ``SubjectData``. If you would like to provide a
+configuration guideline for server owners use ``PermissionDescription``\'s role-templates instead.
+
+.. warning::
+
+    You should think carefully before granting default permissions to users. By granting the permissions you are
+    assuming that all server owners will want these defaults (at least the first time the plugin runs) and that
+    exceptions will require server owners to explicitly deny the permissions (which can't even be done without a custom
+    permissions service implementation). This should roughly correspond to a guest on a single player lan world without
+    cheats. For example a chat plugin would allow sending chat messages by default to imitate vanilla game behaviour
+    for features that were changed by the plugin.
+
+.. note::
+
+    The default ``Subject``\s' persistent ``SubjectData``\s take precedence over the transient ones.
+    For all other ``Subject``\s the transient ``SubjectData``\s take precedence over the persistent ones.
+
+If neither the Subject, nor any of its parents, nor the defaults assign a value to a permission,
+then it is automatically denied.
+
+.. note::
+
+    Order of precedence in descending order:
+    
+    * Subject itself
+        * Transient
+        * Persistent
+    * Parent Subjects
+        * Transient
+        * Persistent
+    * SubjectCollection Defaults
+        * Persistent
+        * Transient
+    * PermissionService Defaults
+        * Persistent
+        * Transient
+    * Deny permission
+
+SubjectCollections
+==================
+
+A container for subjects that can be used to obtain a Subject by name.
+These are the default Subject Collections:
+
+* User
+    * Contains all on-line ``Player``\s and all off-line ``User``\s (at least those with none-default settings).
+* Group 
+    * Contains all group ``Subject``. Groups are a simple way of structuring a ``Subject``\'s inheritance tree using
+      named ``Subject``\s. Groups should be used if a specific subset of ``Subject``\s have additional permission
+      settings such as a team, faction or role.
+* System
+    * Contains other ``Subject``\s used by the server such as the the console and possible remote consoles. 
+* Command Block
+    * Contains all ``Subject``\s for command blocks. These are useful if you would like to run a ``CommandBlock`` only
+      with the permissions of the creator.
+* Role Template
+    * Contains all role template subjects that are used in ``PermissionDescription``\s. Useful to lookup all recommended
+      permissions for a user. These should not be used for inheritance.
+
+.. note::
+
+    When ``SubjectCollection``\s are queried for a ``Subject`` they will automatically be created, if they do not already
+    exist. However they might not necessarily show up in ``getAllSubjects()`` unless none-default values are set.
+
+SubjectData
+===========
+
+SubjectData are the actual permission stores connected to the Subject.
+There are two types of Subject stores:
+
+* Transient = Only lasts for the duration of the session, it is never saved
+* Regular (persistent) = Might be saved somewhere, and therefore be persisted and exist forever. Its recommended for
+  ``PermissionService``\s to implement a persistent store, however it is not a requirement. It might also depend on the
+  subject type. If there is no persistence then the transient store will be returned in both methods.
+
+Plugin authors should consider whether it is necessary to persist a value when choosing between them.
+
+* If it is only for a short time (e.g. during a minigame) then use the transient one.
+* If it is for a long time or forever (e.g. a promotion to VIP) use the regular (persistent) one.
+
+Please refer to the Inheritance section if want to know more about the inheritance and precedence of the transient
+and persistent ``SubjectData``\s.
+
+Subject Options
+===============
+
+Subjects also provide the possibility to store string options. These are basically key value pairs that can be
+assigned and inherited. Unlike the permission strings the keys are not hierarchical and don't provide any inheritance
+mechanisms themselves, but the key value pairs itself are inherited from parent ``Subject``\s in the same way permissions
+are.
+
+Contexts
+========
+
+If you consider each permission to a privilege or ability to be able to do something, a :javadoc:`Context` is the
+circumstances where that privilege is usable.
+
+You might want to give a ``Subject`` permission to do something, but only when the ``Subject`` is in a certain world,
+or in a certain region.
+
+Contexts are accumulated by a ``Subject``, and are then used by the ``PermissionService`` to decide if the ``Subject``
+has a privilege or not.
+
+Sponge provides some contexts by default, but it is generally down to other plugins to provide additional contexts to
+the PermissionService, through a :javadoc:`ContextCalculator`.
+
+When creating contexts for your own plugin please try to avoid conflicts with other plugins (e.g. by prefixing the
+context key with your plugin id) unless these contexts are meant to be shared.
+
+.. note::
+    
+    Please make sure that your ``ContextCalculator`` responds as fast as possible as it will get called frequently.
+
+Example
+~~~~~~~
+
+Your ``ContextCalculator`` may look like this:
+
+.. code-block:: java
+
+    import org.spongepowered.api.command.CommandSource;
+    import org.spongepowered.api.service.context.Context;
+    import org.spongepowered.api.service.context.ContextCalculator;
+    import org.spongepowered.api.service.permission.Subject;
+    import java.util.HashMap;
+    import java.util.Map;
+    import java.util.Set;
+    import java.util.UUID;
+
+    public class ExampleCalculator implements ContextCalculator<Subject> {
+    
+        private static final Context IN_ANY_ARENA = new Context("myArenaPlugin-inAnyArena", "true");
+        private static final Context NOT_ANY_ARENA = new Context("myArenaPlugin-inAnyArena", "false");
+        private static final String ARENA_KEY = "myArenaPlugin-arena";
+    
+        private final Map<UUID, String> playerArenas = new HashMap<>();
+    
+        @Override
+        public void accumulateContexts(Subject calculable, Set<Context> accumulator) {
+            final Optional<CommandSource> commandSource = calculable.getCommandSource();
+    
+            if (commandSource.isPresent() && commandSource.get() instanceof Player) {
+                final Player player = (Player) commandSource.get();
+    
+                final UUID uuid = player.getUniqueId();
+                if (this.playerArenas.containsKey(uuid)) {
+                    accumulator.add(IN_ANY_ARENA);
+                    accumulator.add(new Context(ARENA_KEY, this.playerArenas.get(uuid)));
+                } else {
+                    accumulator.add(NOT_ANY_ARENA);
+                }
+            }
+        }
+    
+        @Override
+        public boolean matches(Context context, Subject subject) {
+            if (!context.equals(IN_ANY_ARENA) && !context.equals(NOT_ANY_ARENA) && !context.getKey().equals(ARENA_KEY)) {
+                return false;
+            }
+    
+            final Optional<CommandSource> commandSource = subject.getCommandSource();
+            if (!commandSource.isPresent() || !(commandSource.get() instanceof Player)) {
+                return false;
+            }
+    
+            final Player player = (Player) commandSource.get();
+    
+            if (context.equals(IN_ANY_ARENA) && !this.playerArenas.containsKey(player.getUniqueId())) {
+                return false;
+            }
+    
+            if (context.equals(NOT_ANY_ARENA) && this.playerArenas.containsKey(player.getUniqueId())) {
+                return false;
+            }
+    
+            if (context.getKey().equals(ARENA_KEY)) {
+                if (!this.playerArenas.containsKey(player.getUniqueId())) {
+                    return false;
+                }
+    
+                if (!this.playerArenas.get(player.getUniqueId()).equals(context.getValue())) {
+                    return false;
+                }
+            }
+    
+            return true;
+        }
+    }
+
+The ``ContextCalculator`` can be registered via: 
+
+.. code-block:: java
+    
+    permissionService.registerContextCalculator(contextCalculator);


source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,183 @@
...
+
+* ``myPlugin.commands.teleport.execute``
+
+Only grants the user the permission to execute the command.
+(With this permission alone the user would only be able to teleport himself to others in his current world.)
+
+* ``myPlugin.commands.teleport.others``
+
+Only grants the user the permission to teleport others.
+(He still needs the ``myPlugin.commands.teleport.execute`` to actually execute the command.)
```
Inscrutable: It's great to have examples, but this gets a bit long - and is bloated by the repetition of this line. Perhaps you can state that it is a prerequisite for the following permissions. However, is it truly necessary to have an .execute permission, or is that just a choice in your (theoretical) implementation?

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,183 @@
...
+Subject
+=======
+
+A Subject is a holder of assigned permissions it can be obtained from all CommandSources.
+Permissions can be granted or denied. Plugin authors will usually only deal with Subjects.
+
+Example
+~~~~~~~
+
+In the permission plugin's config this might look like this.
```
Inscrutable: This sentence is a bit clumsy. Drop the "In" and the first "this" and it makes more sense to me.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,183 @@
...
+    
+    users:
+      Spongie:
+        permissions:
+          - myPlugin.commands.teleport.execute
+          - myPlugin.commands.teleport.others
+          - myPlugin.commands.teleport.worlds
+          - !myPlugin.commands.teleport.worlds.protectedworld
+
+In this case the user `Spongie` was granted the permission to execute the teleport command for himself and other players
```
Inscrutable: Our Spongie is female.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,183 @@
...
+to teleport him to all worlds, but `protectedworld`.
```
Inscrutable: all worlds, other than `protectedworld`.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,183 @@
...
+automatically inherit the ``myPlugin.commands.teleport.all`` permission allowing him to teleport all players at once.
+He still is not able to teleport himself or others to the `protectedworld` though.
+
+If neither the subject itself nor any parent subjects grant or deny a permission then it will be inherited from the
+default subject. The permission plugin/the server owner alone is allowed to define persistent defaults. 
+All other plugins are required to define `transient`/`virtual` defaults for their own permissions.
+
+.. note::
+
+    Persistent permission assignments always take precedence before transient ones.
```
Inscrutable: ... "take precedence over transient ones."

source/plugin/permissions/definitions.rst
ST-DDT: > It's great to have examples, but this gets a bit long

I try to shorten it a bit.

> However, is it truly necessary to have an .execute permission

How else would you grant the user the permission to execute that command?
Of course you could check each sub permission (other, worlds etc, but thats way to much work IMO)

source/plugin/permissions/definitions.rst
ST-DDT: Fixed

source/plugin/permissions/definitions.rst
ST-DDT: `@Spongie` I'm very sorry.

source/plugin/permissions/definitions.rst
ST-DDT: Fixed

source/plugin/permissions/definitions.rst
ST-DDT: Was I really that tired while writing it? Fixed

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
+===========
+Definitions
+===========
+
+Permission
+==========
+
+A permission is a case-sensitive??? hierarchical string key that is used to determine whether a Subject can do a specific action or not.
```
lucko: They are case insensitive. 
https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/service/permission/PermissionDescription.java#L98

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+* "A" - "Z"
+* "a" - "z"
+* "0" - "9"
+* "_"
+* "-"
+* "."
+
+Examples
+~~~~~~~~
+
```
lucko: I think it's worth stressing that these are only examples, not requirements.

It is not absolutely necessary that plugins implement "myplugin.commands", nor is it necessary to have ".execute" or ".all".

They're simply suggestions.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+.. tip::
+
+    You might skip writing descriptions for some parent permission groups such as ``myPlugin.commands.teleport.worlds``
+    or ``myPlugin.commands`` as their meaning can be derived from the permission structure and the defined children
+    alone.
+
+Subject
+=======
+
+A Subject is a holder of assigned permissions it can be obtained from all CommandSources.
```
lucko: Also note that **anything** that should, or can hold permissions is a subject. Subjects are not __just__ command sources.

Additionally, important to know that they can be inherited themselves, or inherit other subjects.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+Permissions can be granted or denied. Plugin authors will usually only deal with Subjects.
+
+Example
+~~~~~~~
+
+The permission plugin's config might look like this.
```
lucko: Poor example IMO and totally unnecessary.

Individual implementations can document their own layouts, it's got nothing to do with Sponge Permissions.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+automatically inherit the ``myPlugin.commands.teleport.all`` permission allowing him to teleport all players at once.
+He still is not able to teleport himself or others to the `protectedworld` though.
+
+If neither the subject itself nor any parent subjects grant or deny a permission then it will be inherited from the
+default subject. The permission plugin/the server owner alone is allowed to define persistent defaults. 
+All other plugins are required to define `transient`/`virtual` defaults for their own permissions.
+
+.. note::
+
+    Persistent permission assignments always take precedence over transient ones.
```
lucko: It's not a requirement. 

For defaults this might make sense, but for regular Subjects, it would make more sense that transient permissions would override permanent ones. 

Either way, it's not documented, nor a requirement. It's totally down to the implementing plugin to decide.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+denied.
+
+.. note::
+
+    Order of precedence in descending order:
+    
+    * Subject himself
+    * Parent subjects
+    * Defaults
+      * Persistent (Server owner via Permission plugin)
```
lucko: Again, that might not be true. It's not a requirement in the JavaDocs.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+==================
+
+A container for subjects that can be used obtained a Subject by name.
+These are the default Subject Collections:
+
+* User/Player
+* Group 
+* System
+* Command Block
+* Role Template
```
lucko: Perhaps actually explain what these are. What goes in a Role Template SubjectCollection? What's it's use? 😕

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+    You might skip writing descriptions for some parent permission groups such as ``myPlugin.commands.teleport.worlds``
+    or ``myPlugin.commands`` as their meaning can be derived from the permission structure and the defined children
+    alone.
+
+Subject
+=======
+
+A Subject is a holder of assigned permissions it can be obtained from all CommandSources.
+Permissions can be granted or denied. Plugin authors will usually only deal with Subjects.
+
```
lucko: I'd also disagree with that. You can't get a Subject without using the main PermissionService and then a SubjectCollection.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+Example
+~~~~~~~
+
+The permission plugin's config might look like this.
+
+.. code-block:: text
+    
+    users:
```
lucko: Additionally this is YAML, which isn't really used by any plugins, afaik.

Most providers currently are storing in json.

Either way, it's pointless documenting it, as it will be of no use at all. Different providers will use different formats.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+          - !myPlugin.commands.teleport.worlds.protectedworld
+
+In this case the user `Spongie` was granted the permission to execute the teleport command and teleport her or other players to all worlds, , other than `protectedworld`.
+
+Inheritance
+~~~~~~~~~~~
+
+Subjects also inherit permissions from their parents, unless the given permission is defined in the subject itself.
+Following that rule if any of his parents is granted the ``myPlugin.commands.teleport`` permission the user would
+automatically inherit the ``myPlugin.commands.teleport.all`` permission allowing him to teleport all players at once.
```
lucko: This bit should ideally go in it's own section IMO.

We should separate "vital" information, about plugin's checking for and providing their own permissions from information about how to use PermissionService.

source/plugin/permissions/definitions.rst
```diff
@@ -0,0 +1,181 @@
...
+
+OptionSubject
+=============
+
+TODO: An inheritable String value such as a chat prefix.
+
+Contexts
+========
+
+TODO: A context (name) that may or may not apply to a `Subject` and can be used to assign permissions only in specific
```
lucko: My best description would be:

If you consider each permission to a privilege or ability to be able to do something, a Context is the circumstances where that privilege is usable.

You might want to give a Subject permission to do something, but only when the Subject is in a certain world, or in a certain Region.

Contexts are accumulated by a Subject, and are then used by the PermissionService to decide if the Subject has a privilege or not. 

Sponge provides some contexts by default, but it is generally down to other plugins to provide additional contexts to the PermissionService, through a ContextCalculator.

Tzky:
To anwer the question you asked on the last paragraph on this PR:
Yes, there is a buildserver for PRs, but sadly it's not available for the public (it's for staff only currently). So if you want to have a preview, you can either use gulp locally or setup your own buildserver+deploy.

ST-DDT:
@Tzky Thanks for the info. I guess I will stick to http://rst.ninjs.org until I get time to setup my own system.

source/plugin/permissions/definitions.rst
ST-DDT: Fixed

source/plugin/permissions/definitions.rst
ST-DDT: Added comment

source/plugin/permissions/definitions.rst
ST-DDT: I hope i fixed this one properly now.

source/plugin/permissions/definitions.rst
ST-DDT: Removed

source/plugin/permissions/definitions.rst
ST-DDT: Removed

source/plugin/permissions/definitions.rst
ST-DDT: Fixed.

source/plugin/permissions/definitions.rst
ST-DDT: Done

source/plugin/permissions/definitions.rst
ST-DDT: Fixed for now. But IMO the written word always have more weight that the transient one.

If a server owner (persistently) defines that a user should not have a specific permission than the transient assignment of a plugin should be ignored.

Could you give me an example in which case it would be better the other way round.

source/plugin/permissions/definitions.rst
ST-DDT: Fixed.

source/plugin/permissions/definitions.rst
lucko: A user usually has a permission set to true, but for the duration of their session, you want to override it with a negated value.

You're forgetting that transient values aren't just for defaults. They might have other uses, where you want to temporarily override a persisting permission.

Usually with permissions, more specific rules would take priority over more generic ones, and I'd consider a transient ruling to be more specific.

Either way, it's not a requirement that providers implement it in a specific way. 😉

source/plugin/permissions/definitions.rst
ST-DDT: Done

source/plugin/permissions/definitions.rst
ST-DDT: Thanks. I guess I will just use this.

ST-DDT:
I updated the definitions. Please check them once more.

Short question in between: What's the use of a `"system"` `Subject`?

Inscrutable:
I'm pretty sure the _system_ subject contains the server console. I managed to set up PEX in a way that denied console commands, early in testing. What use that is, is anybody's guess.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,225 @@
...
+
+Inheritance
+~~~~~~~~~~~
+
+If a ``Subject`` got a permission assigned, it will use that value.
+Otherwise it will be inherited from any parent ``Subject``. It does not matter what kind of parent (ex group or player)
+``Subject`` that might be.
+
+If neither the subject itself nor any parent subjects grant or deny a permission then it will be inherited from the
+default ``Subject``. The default subject can be obtained from the ``PermisionService``.
```
Inscrutable: You left an ``s`` out of ``PermissionService``

source/plugin/permissions.rst
```diff
@@ -2,9 +2,225 @@
...
+    * Contains all ``Subjects`` for command blocks. These are useful if you would like to run a ``CommandBlock`` only
+      with the permissions of the creator.
+* Role Template
+    * Contains all role template subjects that are used in ``PermissionDescriptions``. Useful to lookup all recommended
+      permissions for a user. These should not be used for inheritance.
+
+SubjectData
+===========
+
+SubjectData are the actual permission store connected to the Subject. There are two types of Subject stores. The
```
Inscrutable: I'd use a colon, ie. "Subject stores:".
Additionally, a quibble: should we be saying "saved to disk" or "saved to file"? Maybe it's easier just to say "saved" and stay system agnostic.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,225 @@
...
+transient one, that won't be persistent to the disk, and the persistent one that is loaded from and saved to disk.
+Plugin authors should think whether it is necessary to persist a value when choosing between them. Except for the
```
Inscrutable: think => consider

source/plugin/permissions.rst
ST-DDT: Fixed

source/plugin/permissions.rst
ST-DDT: Changed

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+* "a" - "z"
+* "0" - "9"
+* "_"
+* "-"
+* "."
+
+Inheritance
+~~~~~~~~~~~
+
+If a user has the permission ``myPlugin.commands`` then he automatically has all sub permissions such as
```
lucko: replace "he" --> "they", "has" --> "have"

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+    * Only grants the user the permission to teleport to all worlds.
+
+* ``myPlugin.commands.teleport.worlds.mynetherworld``
+    * Only grants the user the permission to teleport to *mynetherworld*.
+
+PermissionDescription
+=====================
+
+The permission description is an utility that is meant to provide the server owner with details on a certain permission.
+PermissionDescriptions is an optional feature a PermissionService can provide. The creation of PermissionDescriptions
```
lucko: replace "is" --> "are"

Or, replace "PermissionDescriptions" with "A PermissionDescription"

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+does **not** have any impact on whether a permission exists, who has access or what its default value is.
+
+The description consists of:
+
+* the target permission id
+* a Text description
+* and an assigned role
```
lucko: Remove "and". This is a list

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+.. tip::
+
+    You might skip writing descriptions for some parent permission groups such as ``myPlugin.commands.teleport.worlds``
+    or ``myPlugin.commands`` as their meaning can be derived from the permission structure and the defined children
+    alone.
+
+Subject
+=======
+
+A ``Subject`` is a holder of assigned permissions it can be obtained from the ``PermissionService`` via
```
lucko: New sentence needed between "permissions" and "it".

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+    You might skip writing descriptions for some parent permission groups such as ``myPlugin.commands.teleport.worlds``
+    or ``myPlugin.commands`` as their meaning can be derived from the permission structure and the defined children
+    alone.
+
+Subject
+=======
+
+A ``Subject`` is a holder of assigned permissions it can be obtained from the ``PermissionService`` via 
+``SubjectCollections``.
+``CommandSources`` such as ``Players`` are ``Subjects`` by default, but there are many other types of ``Subjects`` as
```
lucko: Remove "as well". Doesn't make grammatical sense.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+well. Anything that has permissions is a Subject even if it just delegates the checks to a contained Subject.
+Permissions can be granted or denied to a Subject. If a permission is neither granted nor denied its setting will be
+inherited. See Inheritance.
+Subjects provide methods to check whether it has a certain permission or not.
```
lucko: Replace "it" --> "they", "has" --> "have"

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+    public canTeleport(Player subject, World targetWorld) {
+        return subject.hasPermission("myPlugin.command.teleport.execute") 
+                && (subject.getWorld() == targetWorld
+                        || subject.hasPermission("myPlugin.command.teleport." + targetWorld.getName()));
+    }
+
+Inheritance
+~~~~~~~~~~~
+
+If a ``Subject`` got a permission assigned, it will use that value.
```
lucko: Replace "got" --> "has"

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+Otherwise it will be inherited from any parent ``Subject``. It does not matter what kind of parent (ex group or player)
+``Subject`` that might be.
+
+If neither the subject itself nor any parent subjects grant or deny a permission then it will be inherited from the
```
lucko: Replace "grant" and "deny" --> "grants" and "denies"

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+default ``Subject``. The default subject can be obtained from the ``PermissionService``.
```
lucko: It can also be obtained from the SubjectCollection.

There are two layers of defaults. The SubjectCollection's defaults are to be considered before the PermissionService's global defaults.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+It is recommended for plugins to define their own permissions to the default's transient ``SubjectData`` during every
```
lucko: Replace "It is recommended for plugins to"  --> "It is recommended that plugins"

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+    * Parent Subjects
+    * Defaults
+        * Persistent
+        * Transient
+    * Deny permission
+
+SubjectCollections
+==================
+
+A container for subjects that can be used obtained a Subject by name.
```
lucko: Replace "used obtained" --> "used to obtain"

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+These are the default Subject Collections:
+
+* User
+    * Contains all on-line and off-line ``User`` / ``Player`` that exist.
```
lucko: Not necessarily. The likelihood is that this will only contain online players.

Some implementations may still persist any changes made, or they might just return a transient dummy subject.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+    * Contains all ``Subjects`` for command blocks. These are useful if you would like to run a ``CommandBlock`` only
+      with the permissions of the creator.
+* Role Template
+    * Contains all role template subjects that are used in ``PermissionDescriptions``. Useful to lookup all recommended
+      permissions for a user. These should not be used for inheritance.
+
+SubjectData
+===========
+
+SubjectData are the actual permission store connected to the Subject.
```
lucko: Replace "store" --> "stores"

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+There are two types of Subject stores:
+
+* The transient one, that won't be persistent at all
+* and the persistent one, that is loaded from and saved to a persistent storage
```
lucko: You're right, but it's not a very clear description IMO. 

Transient = Only lasts for the duration of the session, is never saved
Regular (persistent) = **Might** be saved somewhere, and therefore be persisted and exist forever.

(It is not a requirement that data is persisted.)

Some users might not even understand what "persistent" means, so explaining in terms of the session's duration is probably easier for most users to understand.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,229 @@
...
+
+Plugin authors should consider whether it is necessary to persist a value when choosing between them. Except for the
+default's subject data it is not specified which one takes precedence over the other.
+
+Subject Options
+===============
+
+Subjects may also provide the possibility to store string options. These are basically key value pairs that can be
```
lucko: Remove "may". It is a requirement.

source/plugin/permissions.rst
ST-DDT: Fixed.

source/plugin/permissions.rst
ST-DDT: Changed

source/plugin/permissions.rst
ST-DDT: Fixed

source/plugin/permissions.rst
ST-DDT: Fixed.

source/plugin/permissions.rst
ST-DDT: Fixed

source/plugin/permissions.rst
ST-DDT: Added

source/plugin/permissions.rst
ST-DDT: I changed "default" to "defaults" so I won't change that.

source/plugin/permissions.rst
ST-DDT: Fixed.

source/plugin/permissions.rst
ST-DDT: Fixed.

source/plugin/permissions.rst
ST-DDT: Fixed

source/plugin/permissions.rst
ST-DDT: Fixed.

source/plugin/permissions.rst
ST-DDT: Fixed.

source/plugin/permissions.rst
```diff
@@ -228,14 +228,65 @@ mechanisms themselves.
...
+Your ``ContextCalculator`` may look like this:
+
+.. code-block:: java
+
+    public class MyArenaPlugin implements ContextCalculator<Subject> {
+    
+        private final Context IN_ANY_ARENA = new Context("myArenaPlugin", "*");
+    
+        @Override
+        public void accumulateContexts(Subject subject, Set<Context> accumulator) {
```
ST-DDT: Should I simplify the example to:

````java
@Override
public void accumulateContexts(Subject subject, Set<Context> accumulator) {
    Arena arena = arenasByPlayer.get(subject);
    if (arena != null) {
        accumulator.add(IN_ANY_ARENA);
        accumulator.add(arena.getContext());
    }
}
````
?

Comment (lucko) [inline source/plugin/permissions.rst]: Not a great example IMO. A better example would be to map the name of the arena to the Context, instead of a *, and then only add the context to the accumulator if the user was in an arena.

They're meant to be used as key value pairs, not just a collection of keys.

So:

```
if (user in arena):
  add context where key="myArenaPlugin_arena", value="arenaName"
```

Comment (lucko) [inline source/plugin/permissions.rst]: Ok, better example.

You want a permission to only apply on the "world_nether" world, (I know Sponge gives them DIM-x names, whatever, just an example.)

The key world be "world", the value would be "world_nether". 

There is no such thing as "wildcard" contexts, it's down to the plugin who provides the calculator, or the plugin implementing permissions.

My point is, instead of: key="in_arena", value="*", it would make more sense to have key="in_arena", value="true", or have the value equal the name of the arena.

There's no concept of wildcard Contexts, and especially in the documentation, it's important to show that Contexts are a collection of key value pairs, not a collection of keys.

(I actually think the world example is the easiest to understand, however I understand you're trying to document the calculator here, not the concept of Contexts)

**Comment by ST-DDT** [comment]
@lucko in response to https://github.com/SpongePowered/SpongeDocs/pull/536#discussion_r83534012

> My point is, instead of: key="in_arena", value="*", it would make more sense to have key="in_arena", value="true", or have the value equal the name of the arena.

I changed that. I hope its now easier to understand.

> (I actually think the world example is the easiest to understand, however I understand you're trying to document the calculator here, not the concept of Contexts)

Maybe I can do both.^^

> It's also misleading to have a static Context. The value should vary between Subjects.

Why would they vary? IMO you are either in the arena or not. I could understand adding a context for false too, but per subject?

**Comment by lucko** [comment]
@ST-DDT RE: Static contexts, yeah, you're right. If you've only got a boolean as a value, then yes, I guess a static context is fine, however, as previously mentioned, I think an example where you have more than just two possible values is a better demonstration of the API.

e.g. key=myPlugin_arena, value=some_arena_name

Swapping the "*" for true was my main point, really. Plugins can use Contexts however they like, it's just a case of giving a wide range of appropriate examples so they can understand the point behind them. (and IMO, the current example with just true/false keys doesn't demonstrate Context to the best possible extent. 😄 )

**Comment by ST-DDT** [comment]
Then i think this ready for a final review. Or shall I squash it first?

File: source/plugin/permissions.rst [hunk 36]
```diff
@@ -2,9 +2,293 @@
...
+Example
+~~~~~~~
+
+Your ``ContextCalculator`` may look like this:
+
+.. code-block:: java
+
+    public class MyArenaPlugin implements ContextCalculator<Subject> {
+    
+        private final Context IN_ANY_ARENA = new Context("myArenaPlugin_is_in_arena", "true");
```
kashike: Is this meant to be `static`? Otherwise should not be `ALL_UPPER`.

Inscrutable:
Just waiting on you @kashike then we can pull the lever and make this creation live!

source/plugin/permissions.rst
```diff
@@ -2,9 +2,293 @@
...
+
+* ``myPlugin.commands.teleport.worlds.mynetherworld``
+    * Only grants the user the permission to teleport to *mynetherworld*.
+
+PermissionDescription
+=====================
+
+The permission description is an utility that is meant to provide the server owner with details on a certain permission.
+PermissionDescriptions are an optional feature a PermissionService can provide. The creation of a PermissionDescription
+does **not** have any impact on whether a permission exists, who has access or what its default value is.
```
ryantheleach: Change does to should?

Sure a permissions implementation isn't expected to act based on the permission descriptions currently, but they could, and I don't think it would be outside the spec.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,293 @@
...
+Inheritance
+~~~~~~~~~~~
+
+If a ``Subject`` has a permission assigned, it will use that value.
+Otherwise it will be inherited from any parent ``Subject``. It does not matter what kind of parent (ex group or player)
+``Subject`` that might be.
+
+If neither the subject itself nor any parent subjects grant or deny a permission then it will be inherited from the
+default ``Subjects``. Each ``SubjectCollection`` defines its own defaults. The global and weakest default subject can be
+obtained from the ``PermissionService``. It is recommended that plugins define their own permissions to the default's
```
ryantheleach: Is it really?

It makes sense for plugins with complex context hierarchy, but honestly having default permissions that you need to remove as a server owner is a bigger pain then them having none and explicitly having to grant each one.

There is a reason why we didn't follow Bukkits paths of having permissions be registered and on by default.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,293 @@
...
+
+SubjectCollections
+==================
+
+A container for subjects that can be used to obtain a Subject by name.
+These are the default Subject Collections:
+
+* User
+    * Contains all on-line ``Players``. It may also return subjects for off-line ``Users``. However there aren't any
+      specifications regarding this, so they might be fully transient dummy subjects.
```
ryantheleach: I don't think this is simple enough to understand.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,293 @@
...
+* Group 
+    * Contains all group ``Subject``. Groups are a simple way of structuring a ``Subject's`` inheritance tree.
```
ryantheleach: The main difference between a group and a random subject, is that groups have a name that can be accessed by other plugins.

Groups should be used where it makes sense to have a collection of subjects belonging to a team, faction, or role.

Personally I prefer checking for a permission, but getting all players with a defined permission can be costly if you really do need to be able to get a group, as opposed to checking for a permission.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,293 @@
...
+* System
+    * Contains other ``Subjects`` used by the server such as the the console and possible remote consoles. 
+* Command Block
+    * Contains all ``Subjects`` for command blocks. These are useful if you would like to run a ``CommandBlock`` only
+      with the permissions of the creator.
+* Role Template
+    * Contains all role template subjects that are used in ``PermissionDescriptions``. Useful to lookup all recommended
+      permissions for a user. These should not be used for inheritance.
```
ryantheleach: Why not? I mean, sure they contain the template permissions, but other then that, inheriting off a role should be a decision that the server owner / permissions provider can make.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,293 @@
...
+  ``PermissionServices`` to implement a persistent store, however it is not a requirement. It might also depend on the
+  subject type. If there is no persistence then the transient store will be returned in both methods.
+
+Plugin authors should consider whether it is necessary to persist a value when choosing between them.
+
+* If it is only for a short time (ex during a minigame) then use the transient one.
+* If it is for a long time or forever (ex a promotion to VIP) use the regular (persistent) one.
+
+Except for the default's subject data it is not specified which one takes precedence over the other.
+It is not recommended to set it to both stores at the same time to overcome the unspecified precedence.
```
ryantheleach: You specified the order above, Which is it? Does it have precedence or not?

source/plugin/permissions.rst
```diff
@@ -2,9 +2,293 @@
...
+
+Subject Options
+===============
+
+Subjects also provide the possibility to store string options. These are basically key value pairs that can be
+assigned and inherited. Unlike the permission strings the keys are not hierarchical and don't provide any inheritance
+mechanisms themselves.
```
ryantheleach: I'm not sure this is true.

Options should absolutely be hierarchical and follow the same rules that permissions do.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,293 @@
...
+You might want to give a ``Subject`` permission to do something, but only when the ``Subject`` is in a certain world,
+or in a certain region.
+
+Contexts are accumulated by a ``Subject``, and are then used by the ``PermissionService`` to decide if the ``Subject``
+has a privilege or not.
+
+Sponge provides some contexts by default, but it is generally down to other plugins to provide additional contexts to
+the PermissionService, through a ``ContextCalculator``.
+
+When creating contexts for your own plugin please try to avoid conflicts with other plugins (ex by choosing a unique
```
ryantheleach: Recommend prefixing with plugin id?

source/plugin/permissions.rst
lucko: This is meant to be about documenting the service, not complaining about the way it operates. (I do agree with you though. It was mentioned in a SpongeAPI issue too.)

If you don't register your defaults, then players will not have them by default. It's pretty simple.

Regarding needing to remove them as a server owner, the idea is that plugins register their defaults to the transient SubjectData, and then the persistent SubjectData will override.

source/plugin/permissions.rst
lucko: They're called "templates" for a reason. :wink:

source/plugin/permissions.rst
lucko: I don't see how you could make it any more simple.

- contains users
- guaranteed to contain online users
- might contain offline users too
- if a user is not loaded, a transient dummy (probably) gets returned. changing the state of the returned subject has no impact.

source/plugin/permissions.rst
ST-DDT: I hope the permission plugins will come up with a way to export/dump the transient and persistent permission settings so that you can adjust them as needed.

Also I'm not sure whether the default is required to be in the hierarchy of all Subjects.

lucko: Yes, defaults must be considered. https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/service/permission/Subject.java#L71-L73

ryantheleach: Yes, but I mean, we shouldn't be **recommending** it.

``It is recommended that plugins define their own permissions to the default's``

source/plugin/permissions.rst
lucko: Agreed. That's a much better way to phrasing it too.

source/plugin/permissions.rst
lucko: Well, it's a requirement if an author wants their permission nodes to be given to users by default, but some plugins (e.g. admin tool type plugins) wouldn't want that. Some might also just provide a list of permissions in their documentation. 

So, it's not really a requirement, not a recommendation. Either way you're right, it needs to be clarified.

ST-DDT: Fixed

source/plugin/permissions.rst
ST-DDT: I added some explanations regarding the first part of this.

However for the second one: Getting a group will not help you with the goal of "finding all players with a given permission", because you cannot obtain the children from a ``Subject``. And if you query the inheritance of each player than you can query for the permission directly as well.

source/plugin/permissions.rst
ST-DDT: I hope I fixed that.

source/plugin/permissions.rst
ST-DDT: Fixed.

source/plugin/permissions.rst
```diff
@@ -223,7 +225,8 @@ Subject Options
 
 Subjects also provide the possibility to store string options. These are basically key value pairs that can be
 assigned and inherited. Unlike the permission strings the keys are not hierarchical and don't provide any inheritance
-mechanisms themselves.
+mechanisms themselves. But the key value pairs itself are inherited from parent ``Subjects`` in the same way permissions
+are.
```
ST-DDT: Thx for your fast review.

lucko: :+1:

source/plugin/permissions.rst
```diff
@@ -153,9 +153,9 @@ Otherwise it will be inherited from any parent ``Subject``. It does not matter w
 
 If neither the subject itself nor any parent subjects grant or deny a permission then it will be inherited from the
 default ``Subjects``. Each ``SubjectCollection`` defines its own defaults. The global and weakest default subject can be
-obtained from the ``PermissionService``. It is recommended that plugins define their own permissions to the default's
-transient ``SubjectData`` during every server start-up. This allows the server owner to overwrite the defaults defined
-by plugins according to his needs using the default's persistent ``SubjectData``.
+obtained from the ``PermissionService``. It is recommended (but not mandatory) that plugins define their own permissions
```
ryantheleach: I'm heavily against even having the word recommended even be here.
```
It is possible for plugins to define their own permissions on the default transient ``SubjectData`` during every server start-up, however care should be taken not to introduce unnecessary complexity.

This allows the server owner to overwrite the defaults define by plugins according to their needs using the default's persistent ``SubjectData``.

This however requires the server owner to have knowledge of the default permissions your plugin is granting in the first place, so must be carefully documented.

It is recommended that you do not have any default permissions assigned, unless you are very sure that it will reduce the complexity of your plugin, most plugins should never need to do this.
```

ryantheleach: I'm heavily against even having the word recommended even be here.
```
It is possible for plugins to define their own permissions on the default transient ``SubjectData`` during every server start-up, however care should be taken not to introduce unnecessary complexity.

This allows the server owner to overwrite the defaults defined by plugins according to their needs using the default's persistent ``SubjectData``.

This however requires the server owner to have knowledge of the default permissions your plugin is granting in the first place, so must be carefully documented.

It is recommended that you do not have any default permissions assigned, unless you are very sure that it will reduce the complexity of your plugin, most plugins should never need to do this.
```

lucko: It's a case of whether we want to encourage plugins that work "out of the box" or plugins where some work has to go into configuring their permissions.

I think it's better that you have to explicitly give each permission; it avoids a situation where your players get automatically given access to something you weren't even aware of. It also forces people to read documentation, which is only a good thing.

However, if that's the stance we want to take, then what's the point in having default subjects in the first place?

I'm not really sure how I feel. I can see the benefits and drawbacks of each approach.

ryantheleach: You have seen the default permissions that GP grants right? That's why.

lucko: Eh, the JavaDocs need to be changed then. It's made very clear that PermissionDescriptions should not be used for inheritance or permission calculation. Same with role templates - they're not meant to be used directly.

The API (maybe?) suggests that it would be acceptable to use roles as a template, by copying their SubjectData to a group or the default subjects, but that's really the limit of their use, within the bounds of the JDs.

So to clarify, roles and permission descriptions should **not** be used for default permissions, ever. The API specifically prohibits that.

ryantheleach: I wasn't suggesting that here. 
I was suggesting that giving players permissions by default, as opposed to letting the server admin explicitly define them, with help from the roles.

lucko: Ah, I misunderstood slightly. 

Still, goes back to my previous point:
>However, if that's the stance we want to take, then what's the point in having default subjects in the first place?

ST-DDT: IMO the defaults are useful for the times where you dont have a custom PermissionService or dont want to configure them. Install it and it works.

ST-DDT: Thats why i asked whether the default one is required to be set.
IMHO As long as there is a default its recommended to set it. However I agree with you that the specification needs to be revised to better reflect the actual needs.

lucko: @zml2008 Would you mind giving your opinions here? It's not very clear either way, and I think both sides are pretty valid options. 

In attempt to tl;dr, it's basically: plugins should declare default permissions vs server admins should have to add them manually.

source/plugin/permissions.rst
zml2008: The way PEX implements this is that it treats online and offline players the same -- the only difference is that online players are preloaded into cache on login and offline players may take some time to load. I think it makes sense to mandate that offline players return subjects just as well as online players, since internally they work pretty much the same.

source/plugin/permissions.rst
zml2008: Currently PEX has transient having priority over persistent, but I'm debating whether or not that's the best way to do things, so we should probably figure that out among the permissions plugins before specifying it in the documentation.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,296 @@
...
+
+The permission description is an utility that is meant to provide the server owner with details on a certain permission.
+PermissionDescriptions are an optional feature a PermissionService can provide. The creation of a PermissionDescription
+does **not** have any impact on whether a permission exists, who has access or what its default value is.
+
+The description consists of:
+
+* the target permission id
+* a Text description
+* an assigned role
```
Faithcaio: multiple assigned roles (rename?) are possible or am I wrong?

Inscrutable:
OK, I just want to confirm that everyone is done reviewing, and that we can pull this beast. 
Also, @ST-DDT if you're willing to contribute further docs, we'd be happy to draft you into the SpongeDocs group as well. This is an absolutely stellar effort.

source/plugin/permissions.rst
ryantheleach: This is yet to be resolved / confirmed.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,296 @@
...
+
+If neither the Subject, nor any of its parents, nor the defaults grant and deny a permission, then it is automatically
+denied.
+
+.. note::
+
+    Order of precedence in descending order:
+    
+    * Subject itself
+    * Parent Subjects
```
ryantheleach: Mention transients take precedence over Persistent for subject and parents.

source/plugin/permissions.rst
lucko: I think it was agreed among current plugin authors that for defaults only, having persistent take priority makes the most sense. IMO, this should therefore become a requirement within the API. It's not in the JDs yet, so idk if we want to exclude this section from SpongeDocs until it gets PRed into SpongeAPI.

I'm not sure if zml has made this change for PEX yet, but as far as I know, he plans to.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,296 @@
...
+Inheritance
+~~~~~~~~~~~
+
+If a ``Subject`` has a permission assigned, it will use that value.
+Otherwise it will be inherited from any parent ``Subject``. It does not matter what kind of parent (ex group or player)
+``Subject`` that might be.
+
+If neither the subject itself nor any parent subjects grant or deny a permission then it will be inherited from the
+default ``Subjects``. Each ``SubjectCollection`` defines its own defaults. The global and weakest default subject can be
+obtained from the ``PermissionService``. It is recommended (but not mandatory) that plugins define their own permissions
```
lucko: I don't think it's right to merge this Docs PR until this issue gets resolved. It's a major part of the API. You know my opinions from before, we just need a decision from zml, really.

source/plugin/permissions.rst
lucko: Two assumptions:

* All methods within PermissionService should be "main thread" friendly. I assume that's the same for the rest of Sponge.
* You should absolutely not be doing any sort of I/O, or any blocking operation for that matter, on the main Minecraft server thread.

The two conflict with the idea that SubjectCollection#get should be able to return offline users. So, which of my assumptions is wrong?

ST-DDT:
@Inscrutable

> Also, @ST-DDT if you're willing to contribute further docs, we'd be happy to draft you into the SpongeDocs group as well. This is an absolutely stellar effort.

Thanks for this opportunity. I feel honored. I thought about writing the docs for https://github.com/SpongePowered/SpongeDocs/issues/434 next. Or shall I prioritize others?

source/plugin/permissions.rst
ST-DDT: Has this been decided yet?

lucko: For default subjects: transient takes priority over persistent.
For all other subject types: persistent takes priority over transient.

source/plugin/permissions.rst
zml2008: Yeah, I agree with @ryantheleach here. Allowing permissions by default should not be done in most cases -- primarily if you have a plugin that restricts normal game functionality by permission (like FeatherChat gives chatting a permission, so without defaults installing FeatherChat would block all users from chatting until the server admin configured the plugin), or want to explicitly block permissions from being included when a user grants all permissions (VNP did this on Bukkit iirc).

source/plugin/permissions.rst
Faithcaio: the other way around?
I feel transient has a higher priority than persistent data as its more specific. (except for defaults)

ST-DDT: IMO the server owner should always have the  ability to overwrite any plugin provided permission value.
=> Persistent is always stronger than transient

If a plugin wants to temporarily wants to grant a permission it should do so in an overwriteable/transient way or should use custom contexts calculators.

lucko: That's exactly what has been said already, just in slightly different words. A server owner can override the transient defaults, by adding a default that persists. :wink:

ST-DDT: @lucko Not just the defaults. Every subject's persistent values takes priority over transient ones.
I just want to make sure before changing the docs.

source/plugin/permissions.rst
ST-DDT: Fixed

source/plugin/permissions.rst
ST-DDT: There are quite a few places in the docs that are effected in a similar way.
I will change it to "discouraged" for now and wait for a decision on this topic.

source/plugin/permissions.rst
lucko: No. For all subjects except defaults, transient takes priority over persistent. The reasoning was given above by Faithcaio; transient is considered more specific, and therefore takes priority.

source/plugin/permissions.rst
lucko: Perhaps add the example of using defaults to re-add default game behaviour? I think it's an important point to be made. @zml2008 probably gave the best example of this, referring to use of the chat. Idk, probably would be good to have somewhere on that page?

ST-DDT:
>> Faithcaio: I feel transient has a higher priority than persistent data as its more specific. (except for defaults)

> lucko: No. For all subjects except defaults, transient takes priority over persistent. The reasoning was given above by Faithcaio; transient is considered more specific, and therefore takes priority.

In which way is it more specific?

Lets use a real world example.

1. The server owner is the highest instance.
  - `Boss A` is the top-most boss of a company called "server"
2. The server owner (can) defines a written law: **User** X may not do Y.
  - `Boss A` defines employee `User X` is not allowed to go to `Secure Room Y`.
3. A plugin Z wants to temporarily allow X to do Y, but it can't, because it is only allowed to act in conjunction with the server owner's will.
  - The `Team leader Z` wants `User X` to temporarily help in `Secure Room Y`. It can't be helped because `Boss A` had said: No. Wanna change that? Ask the `Boss A` to allow it in this context or reconsider the No.

Or am I wrong here at any point?

lucko:
That's a pretty weird example IMO. (maybe I'm just too stupid to understand)

The basic idea is that a transient permission will not last for as long as a persistent one, and therefore is more "temporary". As it's more "temporary" than a standard permission that just sits in storage, it's therefore more specific.

My example of that is: 
As a server owner, you have a plugin which stops users from breaking blocks. They get granted a negated permission for "somePlugin.build", however, just for the current session, you want a user to be able to break blocks to they can make changes to the map. You add a transient permission to override the normal blocking behaviour.

I think the actual question we should be asking is whether transient/persistent nodes are flattened before individual node inheritance is applied.

```
somePlugin=false (transient)
somePlugin.something.allowbuild=true (persistent)
```
Assuming that transient does in fact take priority, which is more specific in this case? I know it's certainly not documented in the API.

Maybe for the sake of this PR, and @ST-DDT's sanity, I shouldn't have even asked. Maybe let's ignore that for now. 😆

ryantheleach:
I think permissions providers have been incorrectly assuming that somePlugin.something would be more specific.

Some world rules for arena building.
somePlugin.somthing.allowbuild.grass = true
somePlugin.somthing.allowbuild.tnt = false


A transient permission for the round while the arena is in play.
somePlugin.something.allowbuild = false.

This example it's clear what the plugins and by proxy server owner had in mind. and that is that all building should be denied for the duration of the arena session.

Doing otherwise assumes that the arena plugin knows there are sub-nodes, how to get them, and how to iterate over all of them in order to blanket deny / allow something temporarily. 


That said, I'm sure with some effort we can come up with a counter-example to this as well...

Faithcaio:
> A plugin Z wants to temporarily allow X to do Y, but it can't, because it is only allowed to act in conjunction with the server owner's will. 

Why would plugin Z want to do that if not because someone with enough permissions(the server owner) asked it to?
Also if the plugin is evil it could just set the permission on the persistent SubjectData and override anything a server owner had configured.

If a subject has `somePlugin.something: false` and `somePlugin.somthing.allowbuild.grass: true`
on the same subject in the same SubjectData (both transient or persistent) then of course `somePlugin.somthing.allowbuild.grass` is more specific and will get checked first.
If one of them is transient it is always more specific than the persistent one and will get checked first.

I still feel that for normal Subjects (maybe defaults too) the transient data should have higher priority.
That way you can override any permission temporarily (temporarily is the more specific part here).
You can always override transient data by assigning the data earlier on the inheritance tree.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,310 @@
...
+PermissionService's task to respect the permission and subject inheritance.
+
+Example
+~~~~~~~
+
+The following example could be used to check whether the Player is allowed to execute the teleport command.
+
+.. code-block:: java
+    
+    public canTeleport(Player subject, World targetWorld) {
```
lucko: Urr, that's not a proper method declaration. I think you're missing `boolean`

source/plugin/permissions.rst
```diff
@@ -2,9 +2,310 @@
...
+* Command Block
+    * Contains all ``Subjects`` for command blocks. These are useful if you would like to run a ``CommandBlock`` only
+      with the permissions of the creator.
+* Role Template
+    * Contains all role template subjects that are used in ``PermissionDescriptions``. Useful to lookup all recommended
+      permissions for a user. These should not be used for inheritance.
+
+.. note::
+
+    If ``SubjectCollections are queried for a ``Subject`` it will automatically be created, if it does not already
```
lucko: `If` --> `When`
`it will automatically be created, if it does not` --> `they will automatically be created, if they do not`

lucko:
I think stuff being discussed about priority ordering (which priority takes priority… :c) is probably best discussed over at SpongeAPI. Even if a conclusive decision is made about it, it’s not anywhere in the Java documentation, so I think you just have to assume that implementations can decide either way - in which case, any discussions here are pointless.

Just taking a slight step back; @ST-DDT has done a great job at explaining above and beyond what was already explained in the JDs. The issues being raised aren’t really about the documentation - rather with the API itself. IMO, it’s probably best to take these discussions to SpongeAPI & allow the docs to be merged. I know zml has started on a 6.x branch, hopefully the remaining ambiguities will be resolved then. 🙂 

IMHO this has probably gone on long enough now (hypocritical I know, I’ve definitely been a main source of problems/issues - sorry!), I think it’s probably ready. The remaining issues can be ironed out over time, as and when the appropriate changes are made to the API side of things. 😄

source/plugin/permissions.rst
ST-DDT: Ups, fixed.

source/plugin/permissions.rst
ST-DDT: Fixed

ST-DDT:
Minor question regarding https://github.com/SpongePowered/SpongeDocs/issues/541
Is it okay to store hard references to `Subject`s?
How do I compare subjects? => Is the usage of `equals()` required?

lucko:
The API doesn't really mention anything about implementations of `#equals`.

I would guess that most implementations wouldn't allow more than once instance of the same Subject, in which case, comparing a Subject's identifier and the name of it's containing Collection is probably the most reliable approach. I would probably discourage storing any reference to Subjects, especially in Sets and Maps. You're relying on implementations implementing equals & hashcode properly - but, that's just my personal opinion. I guess it can just be left out until now. (and no, as far as I know, the issue raised in #541 does not apply to anything in PS.)

ST-DDT:
@lucko I ask because `Player` implements `Subject` (=>Do not hard ref), but will the `SubjectCollection` return `Player` instances? 
* If yes (always), than it will be a dummy player sometimes (+bad sideeffects?)
* If yes (when online), then the comparison using equals will not work for freshly (dis-)connected subjects
* If no, then the comparion using equals with online players will not work.

But I guess this is something for the API to specify...

lucko:
No. SubjectCollections will return raw subjects (the implementation's class that just `implements Subject`)

Players get wrapped with the Subject provided by the SubjectCollection.
https://github.com/SpongePowered/SpongeCommon/blob/bleeding/src/main/java/org/spongepowered/common/mixin/core/command/MixinSubject.java

SubjectCollection should never return Player instances. (you'd actually cause StackOverflowExceptions doing so.)

ST-DDT:
Oh yeah, you are right. I thought that you can check with equals whether a Subject belongs to a Player, but this is not possible. So basically a Set<Subject> is useless/not useable in most cases.

lucko:
Sure, although actually, that's a good point to be made about ContextCalculators. 

Simply checking `instanceof Player` isn't enough to determine whether a Subject is a Player.

ST-DDT:
Updated the precedence parts for `SubjectData` according to our discussion on IRC yesterday.

ST-DDT:
I guess its ready for a final review and merge.

I excluded the parts specified in https://github.com/SpongePowered/SpongeAPI/pull/1411 for now.
These can be added in a later PR.

source/plugin/permissions.rst
```diff
@@ -231,8 +235,8 @@ Plugin authors should consider whether it is necessary to persist a value when c
 * If it is only for a short time (ex during a minigame) then use the transient one.
```
Inscrutable: ex should be e.g. (exempli gratia, "for example"). Latin and all that jazz.

source/plugin/permissions.rst
```diff
@@ -231,8 +235,8 @@ Plugin authors should consider whether it is necessary to persist a value when c
...
 * If it is for a long time or forever (ex a promotion to VIP) use the regular (persistent) one.
 
-Except for the default's subject data it is not specified which one takes precedence over the other.
-It is discouraged to set it to both stores at the same time to overcome the unspecified precedence.
+Please refer to the Inheritance chapter if want to know more about the inheritance and precedence of the transient
```
Inscrutable: It gets a whole *chapter*? I hope you mean *page* (or docs).

ST-DDT: Is the term section more appropriate?
Or maybe paragraph?

Inscrutable: Section will do.

source/plugin/permissions.rst
```diff
@@ -169,8 +169,8 @@ guideline for server owners use ``PermissionDescription's`` role-templates inste
 
 .. note::
 
-    Its only specified for the default ``Subjects`` that the persistent ``SubjectData`` takes precedence over transient
-    ones.
+    The default ``Subjects'`` persistent ``SubjectDatas`` take precedence over the transient ones.
+    For all other ``Subjects`` the transient ``SubjectDatas`` take precedence over the persistent ones.
 
 If neither the Subject, nor any of its parents, nor the defaults grant and deny a permission, then it is automatically
```
ST-DDT: Fixed. I hope.

source/plugin/permissions.rst
ST-DDT: Fixed

source/plugin/permissions.rst
ST-DDT: Fixed

ST-DDT:
@Inscrutable Are the imports fine like this?

ST-DDT:
Has anybody some more change requests? If not then I will squash and merge it tomorrow evening.

source/plugin/permissions.rst
```diff
@@ -2,9 +2,371 @@
...
+    * Only grants the user the permission to teleport to all worlds.
+
+* ``myPlugin.commands.teleport.worlds.mynetherworld``
+    * Only grants the user the permission to teleport to *mynetherworld*.
+
+PermissionDescription
+=====================
+
+The :javadoc:`PermissionDescription` is an utility that is meant to provide the server owner with details on a certain
+permission. ``PermissionDescriptions`` are an optional feature a :javadoc:`PermissionService` can provide. The creation
```
ST-DDT: Yesterday I had issues with that because the tag needs to be surrounded by whitespace, but I will try it again.

ST-DDT: @12AwsomeMan34 I might have got you wrong
Are you referring to section headline or the first sentence in those sections?

PR #539, opened 2016-10-25
5.1 / 6.0.0 chunk-gc-tick-interval config change.

ryantheleach:
```
04:22   <@blood|wrk>    whoever updates our config docs
04:22   <@blood|wrk>    change the following
04:22   <@blood|wrk>    chunk-gc-tick-interval should be default 600 now (30 seconds)
04:22   <@blood|wrk>    chunk-unload-delay 15
04:22   <@blood|wrk>    not sure if the defaults are on docs but if so, change it to above
09:42   <+ryantheleach> blood as of which version? target api5 docs or 6 docs?
09:44   <+Deamon>   ryantheleach, it'll end up in 6.0.0 and 5.1.0 I expect
```

### DIFF
--- a/source/server/getting-started/configuration/sponge-conf.rst
+++ b/source/server/getting-started/configuration/sponge-conf.rst
@@ -254,10 +254,10 @@ was generated using SpongeForge build 1399, SpongeAPI version 4.1:
             chunk-gc-load-threshold=0
             # The tick interval used to cleanup all inactive chunks in a world. 
             # Set to 0 to disable which restores vanilla handling. (Default: 1)
-            chunk-gc-tick-interval=1
+            chunk-gc-tick-interval=600
             # The number of seconds to delay a chunk unload once marked inactive. (Default: 30)
             # Note: This gets reset if the chunk becomes active again.
-            chunk-unload-delay=30
+            chunk-unload-delay=15
             # If enabled, any request for a chunk not currently loaded will be denied (exceptions apply for things like world gen and player movement). 
             # Note: As this is an experimental setting for performance gain, if you encounter any issues then we recommend disabling it.
             deny-chunk-requests=true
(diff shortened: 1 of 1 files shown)


source/server/getting-started/configuration/sponge-conf.rst
```diff
@@ -254,10 +254,10 @@ was generated using SpongeForge build 1399, SpongeAPI version 4.1:
             chunk-gc-load-threshold=0
             # The tick interval used to cleanup all inactive chunks in a world. 
             # Set to 0 to disable which restores vanilla handling. (Default: 1)
```
Inscrutable: That's what we usually do (although the entity lists self-populate as they are encountered so they can be a pain). We'll generate one when it appears in an implementation.

12AwsomeMan34: Been a while since I updated sponge, but last I checked there's a config option that you have to turn on to populate the entities list. Otherwise, it defaults to the group's setting, i.e. ambient, aquatic, etc.

12AwsomeMan34:
As @Deamon5550 said, it's probably best to let the server generate a default config and using that instead of updating it by hand. That said, I have a rather old utility called [SpongeConfCompare](https://github.com/12AwsomeMan34/Sponge-ConfCompare) that shows the differences between the what's on the docs and a local config file. It's easier than trying to see what's changed by hand, and you can't really copy a new config onto the docs page since the table below would need to be updated as well.

Inscrutable:
So, the choices @ryantheleach : generate a fresh new config and add that, use the tool above to compare the diff, or just fix the default value line by hand (which is what this issue was about) and let someone worry about the rest in a new PR. Your choice ;)
Is this change in API 5.1.0 now, @Deamon5550 ?

Inscrutable:
@ryantheleach This PR seems to have picked up some stray commits from the past. Some cleanup may be advised, but we want to get this in soon. Also bump @Deamon5550 ^

Deamon5550:
@Inscrutable Yes this is in 5.1.0

PR #549, opened 2016-11-10
Fixed missing parentheses in lambda function.

bibo38:
Since it's a function, it needs parentheses :wink: .

### DIFF
--- a/source/plugin/optional/usage.rst
+++ b/source/plugin/optional/usage.rst
@@ -205,7 +205,7 @@ just use
 .. code-block:: java
 
     getOptionalString()
-          .map(s -> s.toLowerCase)
+          .map(s -> s.toLowerCase())
           .ifPresent(myPlugin::doSomethingWithString);
 
 .. tip::


source/plugin/optional/usage.rst
```diff
@@ -205,7 +205,7 @@ just use
 .. code-block:: java
 
     getOptionalString()
-          .map(s -> s.toLowerCase)
+          .map(s -> s.toLowerCase())
```
Minecrell: This can be just `.map(String::toLowerCase)` to make it even simpler

ST-DDT: I'm not sure but maybe this "longish" version is here on purpose to show both variants. 
The next line is the short version.

Minecrell: Then we should find an example that can't be written as method reference ;)

Inscrutable: If there aren't any more quibbles we should merge this ASAP.

PR #558, opened 2016-12-19
Document some bad practices that should be avoided: Fixes #541

### LINKED ISSUE #541: NEVER Store direct references to entities in Sets/Maps in plugins
Referring to [this important PSA](https://forums.spongepowered.org/t/do-not-store-player-entity-references-in-maps-sets-directly/15513) we need to document this to prevent plugins causing memory leaks. This could get a mention in "Best Practices" and/or a relevant Data page.

> <blood_> https://github.com/SpongePowered/SpongeAPI/commit/167fe7b185bdba8e4addb4ced27d72c60da84c42
> <blood_> this needs to be documented
> <blood_> more info can be found here
> <blood_> https://forums.spongepowered.org/t/do-not-store-player-entity-references-in-maps-sets-directly/15513/10
> <blood_> as well as my respective commits
> <blood_> We now do exactly what Vanilla/Forge do on respawns
> <blood_> and this fixes major compatibility issues with a bunch of mods

ST-DDT:
Documents the following issues new developers might make.

* Storing References (https://github.com/SpongePowered/SpongeDocs/issues/541)
* IO on the main thread
* Accessing game objects outside the main thread


Spongy:

Mumfrey:
> Accessing game objects outside the main thread

This isn't "bad practice", it needs to be absolutely front and centre not tucked away in a bad practices section. It should be included as a fundamental point when discussing both the scheduler (it probably already is) and event listeners.

source/plugin/practices.rst
```diff
@@ -61,3 +70,59 @@ Plugin Interoperability
 =======================
 
 An explanation of how to communicate with other plugins, *TBA*.
+
+
+Bad Practices
+=============
+
+These should practices should be avoided. As they lead to memory leaks (``OutOfMemoryError``), lags or inconsistencies.
```
Inscrutable: Make this one sentence (...avoided, as ...).
They *can* lead to memory leaks - unless you're 100% certain they will?
Lag does not need to be a plural.

source/plugin/practices.rst
```diff
@@ -61,3 +70,59 @@ Plugin Interoperability
...
+* :javadoc:`CommandSource`
+* :javadoc:`Entity`
+* :javadoc:`Player`
+* :javadoc:`Subject`
+* :javadoc:`World`
+* and containers that **MIGHT** contain any of the above elements, including
+    * ``Collections``
+    * ``Maps``
+
+should **NEVER** be stored or be cached in plugins.
```
Inscrutable: The second ``be`` is redundant.

source/plugin/practices.rst
```diff
@@ -61,3 +70,59 @@ Plugin Interoperability
...
+
+These are the main reasons for this:
+
+* The references prevent proper garbage collection
+* The instances might no longer be valid
+
+This can easily be avoided by using the corresponding snapshots or saving the UUID of given the instances and request a
```
Inscrutable: To keep tense consistent, use "requesting" instead of "request".

source/plugin/practices.rst
```diff
@@ -61,3 +70,59 @@ Plugin Interoperability
...
+live instance when you need it.
+
+
+IO on the main thread
+~~~~~~~~~~~~~~~~~~~~~
+
+Executing some IO operations such as loading a config/data file or checking for updates/connecting to a website takes
+much time and greatly affects the TPS on the server. Such tasks should be either done in their own threads or using the
+inbuilt scheduller's async feature.
```
Inscrutable: Typo. It's scheduler

source/plugin/practices.rst
```diff
@@ -61,3 +70,59 @@ Plugin Interoperability
...
+
+This can easily be avoided by using the corresponding snapshots or saving the UUID of given the instances and request a
+live instance when you need it.
+
+
+IO on the main thread
+~~~~~~~~~~~~~~~~~~~~~
+
+Executing some IO operations such as loading a config/data file or checking for updates/connecting to a website takes
+much time and greatly affects the TPS on the server. Such tasks should be either done in their own threads or using the
```
Inscrutable: Move ``either`` ie. "Such tasks should be done either in their own thread or ..."
I'd also like to see a comma after "threads".

source/plugin/practices.rst
ST-DDT: Fixed.

source/plugin/practices.rst
ST-DDT: Done

source/plugin/practices.rst
ST-DDT: Done.

source/plugin/practices.rst
ST-DDT: My most favorite mistake...Fixed.

ST-DDT:
@Inscrutable Thanks for reviewing all my PRs.

ST-DDT:
@Mumfrey I made the warning stronger and also added it to the scheduler docs.

Is this sufficient or shall I change it some more.

IMO its a good advice to have that warning in the best practise section as well or shall I remove it from there to reduce redundancies? (I recommend using async for lengthy tasks a few lines above so it might be useful to also outline the limitations there)

MarkL4YG:
Reading the plugins config on a different thread?
I do think that's a bit overpowered since reading configs should be done after initialization and reloading during operation itself is already considered bad practice - The plugin should simply disallow it entirely or warn the administrator about that.
I also think there's no point in detaching from the main thread during initialization since the server shouldn't be operating while plugins (may) still initialize.

Inscrutable:
I'm hoping someone (else) can review this dusty PR so we can merge it or bury it.

ST-DDT:
@MarkL4YG I added a comment concerning the IO during the server startup. Do you consider this enough?
If yes, then I'm going to merge this.

@Inscrutable Objections?

MarkL4YG:
@ST-DDT Yes, I think that'll do.

Be aware that in the preview the page is still named "Best practices" while the subsection headline "Bad practices" is not really showy compared to the other titles.

ST-DDT:
Should I move it to its own page? Or is there a way to increase the increase the showiness?

MarkL4YG:
Maybe order it the way "Setting up your project" is ordered and call the page "Practices" or sth like that.

source/plugin/practices.rst
```diff
@@ -119,3 +128,68 @@ Plugin Interoperability
 =======================
 
 An explanation of how to communicate with other plugins, *TBA*.
+
+
+Bad Practices
+=============
+
+These should practices should be avoided, as they can lead to memory leaks (``OutOfMemoryError``), lag or
```
Inscrutable: too many "should"s.

source/plugin/practices.rst
```diff
@@ -119,3 +128,68 @@ Plugin Interoperability
...
+This can easily be avoided by using the corresponding snapshots or saving the UUID of the given instances and requesting
+a live instance when you need it.
+
+
+IO on the main thread
+~~~~~~~~~~~~~~~~~~~~~
+
+Executing some IO operations such as loading a config/data file or checking for updates/connecting to a website takes
+much time and greatly affects the TPS on the server. Such tasks should be done either in their own threads, or using the
+inbuilt scheduler's async feature. It is perfecty fine to load required files/config on the main thread during server
```
Inscrutable: "perfectly". It might also be better to phrase this starting with "However, ", and omit the "though" at the end.

source/plugin/practices.rst
```diff
@@ -119,3 +128,68 @@ Plugin Interoperability
...
+startup/plugin initialization though.
+
+.. code-block:: text
+
+    this.game.getScheduler().createAsyncExecutor(this).execute(this::checkForUpdates);
+
+For more details refer to the :doc:`scheduler` docs.
+
+If this is done wrong, you will see clients timing out or the server's watchdog might kill the server.
```
Inscrutable: I suggest "If this is done incorrectly, you may see clients time out, or the watchdog may even kill the server."

source/plugin/practices.rst
```diff
@@ -119,3 +128,68 @@ Plugin Interoperability
...
+
+
+Accessing game objects outside the main thread
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Accessing game objects outside of the main thread can lead to crashes, inconsistencies and various other problems and
+should be avoided. If you have a lengthy operation on a different thread use the :doc:`scheduler` to execute the apply
+part on the main thread. If you want to use an game object in a different thread use a snapshot of the instance or a
```
Inscrutable: "a game object"

dualspiral: "to execute the apply part" -> something like "to make changes to such game objects" (it's not clear what the "apply part" means here)

ST-DDT:
@Inscrutable @dualspiral @MarkL4YG Fixed and split into two pages for better overview.

source/plugin/practices/index.rst
```diff
@@ -0,0 +1,12 @@
+=========
+Practices
+=========
+
+These section describes some best practices that should be followed and some bad practices that should be avoided.
```
Inscrutable: "This section"

ST-DDT: Fixed

ST-DDT:
More changes needed?

PR #560, opened 2017-01-22
Switch to new theme based on sphinx_rtd_theme

Minecrell:
This is a potential new theme for SpongeDocs based on the [Read the Docs Sphinx Theme](https://github.com/snide/sphinx_rtd_theme), with some modifications to make it look good for Sponge (and to add our dropdown menu).

The reason for changing the theme is that our custom theme has increasingly become outdated and more difficult to maintain. To make that easier we would have to rewrite it using Bootstrap. This is a lot of work (and we have only limited time), so using a complete theme that looks good the way it is might be a better option.

Full demo is available at https://sponge.dev.minecrell.net

screenshot_20170122_195459

### TODO
- [x] Landing page
- [x] Add version/language selection
- [x] Add Google Analytics ID


Spongy:

Tzky:
I discussed this with minecrell. Basically our options are:
1. keep the current *non responsive* docs theme
2. port the current theme to bootstrap/bootswatch
3. use one of two sphinx bootstrap themes and make adjustements to fit our docs
4. go with rtd theme and only make *very minor* adjustments, it's responsive by default
5. built a responsive sphinx theme from scratch

As already pointed out above, the current theme is a mess (codewise) and outdated. That leaves us with option 3, 4 and 5. Seeing that we're all very short on time, solution 5 is not an option. Likewise option 3, as there are major changes needed to make the two themes fit. 

TL:DR we should go with option 4 or keep it as it is, as long as we're not satisfied with the result of this PR.

Further thoughts?

Inscrutable:
So it looks like we'll lose some minor formatting tweaks, and also the light/dark BG options for code blocks. We can probably live with that, as long as the other stuff remains accessible. In the longer run, we could always fork one of the bootstrap themes and work on that.

ST-DDT:
The new design looks fine for me.
For the light/dark code blocks....we should decide on one and use that always.

Minecrell:
Don't worry about the code block themes, I've added an option to change them. The dark theme is the same from the old theme, the light one comes from the theme and is now the default (I prefer the light one in this case actually).

Meronat:
I think the plus and minus signs when you open a category on the left are either kind of ugly. I think they should just be removed as it is intuitive where you click or at least changed.

image

image

Minecrell:
I'm unsure about what @Meronat said above, so please like his comment if you'd like to have this change.

kashike:
I like the :heavy_plus_sign: and :heavy_minus_sign: signs.

Tzky:
What about using fa-caret-right (collapsed) and fa-caret-down (expanded) instead?

Minecrell:
A full demo page (with all translated docs) is now available here: https://sponge.dev.minecrell.net/master/en/

The homepage is still missing but switching between the languages should already work using the bottom bar. If you switch languages it should now stay on the same page instead of jumping back to the main documentation page. Please note that switching between versions won't work because the older versions are not getting built currently.

Additionally there is now a new "special" Docs page that allows contributing translations live on the Docs page so you immediately have a preview of how the translated page would look like.* This is implemented using [Crowdin's In-Context Localization](https://crowdin.com/page/in-context-localization).

You can access a demo page here: https://sponge.dev.minecrell.net/master/translate/



\* It doesn't convert links to other pages currently.

Minecrell:
This is mostly ready for merge, I need to finish some last tests but the main work is done.

There is also a new homepage for the Docs now: https://sponge.dev.minecrell.net

screen shot 2017-01-30 at 17 51 55

Let me know if you have any suggestions how to improve it.

ST-DDT:
It might just be my browser but on the landing page (with the flags) the search button does not work (only "enter" does). But there are no results shown even if i press enter. The Search just keeps spinning /dot-ing.

Minecrell:
@ST-DDT
 Known problem, this is a bug in sphinx_rtd_theme that was already fixed, however there isn't a new release yet.

Minecrell:
This is ready for merge now. I think we have agreed to switch to the modified version of the ReadTheDocs theme, at least until we have enough time to work on a completely new theme. We can merge it this weekend if there is nothing else that needs to be changed.

PR #562, opened 2017-01-24
Cut 2016 release date

Inscrutable:
This should also get cherry picked into earlier Docs versions.
When or if we have a new target date for "release" we can revisit this.


Spongy:

PR #572, opened 2017-02-01
Mixins and Hybrid Mods

### LINKED ISSUE #566: Best Practices incorrectly states that plugins cannot use mixins
^ title.
https://github.com/SpongePowered/SpongeDocs/blob/master/source/plugin/practices.rst#mixins

Inscrutable:
Resolves #566

### DIFF
--- a/source/plugin/practices.rst
+++ b/source/plugin/practices.rst
@@ -27,13 +27,6 @@ Economy API
 The Economy API is used to link economy plugins with other plugins that use the economy (i.e. shops). You can read
 about the Economy API :doc:`here <economy/index>`, which details everything you need to know about the API.
 
-Mixins
-~~~~~~
-
-Mixins are specifically for transforming classes before other mods/plugins start. ForgeModLoader calls these mods
-“Coremods”. SpongeForge is a Coremod, and deploys mixins on startup. It is possible for plugins to use mixins, but
-by that point they are better classified as mods.
-
 
 Packets
 ~~~~~~~
@@ -56,6 +49,70 @@ modules to interact with the underlying implementation (SpongeForge or SpongeVan
 
 Plugins using implementation-specific code are very likely to break between versions, and should be clearly labelled
 as such wherever they are hosted. These may more appropriately labelled as "Mods".
+
+
+Mixins
+~~~~~~
+
+Mixins are specifically for transforming classes before other mods/plugins start. ForgeModLoader calls these mods
+“Coremods”. SpongeForge is a Coremod, and deploys mixins on startup. Mixins can be used by plugins, but be aware of the
+additional complexities involved. 
+
+**Hybrid Mods**
+
+Sponge plugins which leverage mixins may also be a considered core mods, based on content.
+
+- To use mixins in FML, it must be a coremod. The jar may also contain a Sponge plugin,
+  so most properly the container is a "hybrid mod".
+- To use mixins in SpongeVanilla, intentions must be declared in the manifest.
+  SpongeVanilla then injects the mixins.
+- Having both types in a single jar is entirely possible. (Indeed, a single jar could
+  easily contain a tweaker, FML mod, coremod, bukkit plugin, sponge plugin, and/or litemod.)
+  
+Some definitions may be helpful here. 
+
+Tweak Mod (aka Tweaker)
+  a subsystem-level mod which hooks directly into the game using LaunchWrapper, usually used for
+  ModSystems (eg. LiteLoader, FML) and stand-alone mods (eg. Optifine). Can interact with any aspect
+  of the game environment directly. Generally breaks every version.
+
+Core Mod
+  has almost equivalent power of a Tweak Mod but must be bootstrapped by a ModSystem.
+  Can interact with any aspect of the game environment directly. Generally breaks on every new Game version.
+
+Mod
+  interacts with the game only via a ModSystem, the mod is exposed to game objects directly but will
+  generally only hook into the game via hooks provided by the ModSystem. Generally breaks every major
+  version of the Game (depending on features used). The term mod is also used as an umbrella term for
+  anything which modifies the game, though for the sake of clarity we'll use this definition.
+
+Plugin
+  interacts with the game only via an API, does not interact with game objects directly in any way,
+  only leverages objects exposed by the API. Generally breaks only when the API is revised
+  (and sometimes not even then).
+
+It's also important to distinguish the container from the contents. Issues with terminology tends to
+arise because a jar containing a mod tends to get referred to as a "mod".
+Any plugin which is not fully decoupled via the API puts itself into the category of Mod.
+This type of "plugin" may be prevalent where there are shortcomings in an API.
+
+**Advantages of Hybrid Mods**
+
+A hybrid mod leverages both a plugin component which interacts via the API, and a mod (or even coremod)
+in the same package. This has the disadvantages of a mod (breaks every version) but also the power of a
+mod (can interact with the game directly) coupled with some of the benefits of a Plugin (high-level
+abstract access to the game, and can also interact with other plugins as a peer).
+
+The primary benefit of this system is that the maintenance burden is reduced when updating the mod,
+because any features accessed via the API are likely to be much more stable.
+
+This type of mod can be used to implement plugins whose needs overflow the capability of the API (in
+the case of a plugin which needs to leverage mixins for a particular feature); but can also be used
+for mods which want to leverage services afforded by the API (eg. a mod which wants to provide direct
+support for permissions or chat channels).
+
+Unlike NMS-exploiting "plugins", a hybrid mod makes its' nature clear.
+
 
 
 Plugin Interoperability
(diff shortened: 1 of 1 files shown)


Spongy:

Inscrutable:
I'm just wondering if there's more to be said here - as I understand it, adding mixins to a plugin has consequences, such as game version, and having to handle both Sponge implementations.
But I could be wrong.

Mumfrey:
Well let's clear this up by baking some definitions, because it gets very muddy very quick and the statement you're making wanders into this area. My intention is to round out your understanding so that you can decided how to phrase it appropriately:

**Definitions:**
* **Tweak Mod** (aka Tweaker) - a subsystem-level mod which hooks directly into the game using `LaunchWrapper`, usually used for *ModSystems* (eg. LiteLoader, FML) and stand-alone mods (Optifine). Can interact with any aspect of the game environment directly. Generally breaks every version.

* **Core Mod** - has almost equivalent power of a Tweak Mod but must be bootstrapped by a *ModSystem*. Can interact with any aspect of the game environment directly. Generally breaks on every new Game version.

* **Mod** - interacts with the game only via a *ModSystem*, the mod is exposed to game objects directly but will generally only hook into the game via hooks provided by the *ModSystem*. Generally breaks every *major* version of the Game (depending on features used). The term *mod* is also used as an umbrella term for anything which *modifies the game*, though for the sake of clarity we'll use this definition.

* **Plugin** - interacts with the game only via an *API*, does not interact with game objects directly in any way, only leverages objects exposed by the *API*. Generally breaks only when the *API* is revised (and sometimes not even then).

It's important to distinguish the *container* from the *contents*. The issue with terminology tends to arise because a jar containing a mod tends to get referred to as a "mod". Now that jar might contain a coremod, a tweaker, and a mod. Hell in the case of Optifine the jar actually contains both a Tweaker and an installer. Here are some real-world examples:

**Real-world:**
<table width="100%">
    <tr><th width="20%">Thing</th><th width="20%">Categories</th><th width="60%">Notes</th></tr>
    <tr><td>FML, LiteLoader, SpongeVanilla</td><td><ul><li>Tweaker</li></ul></td>
        <td>These Tweakers are <i>ModSystems</i> because their duty includes loading other mods (in fact FML and LiteLoader both adhere to an implicit contract which actually means they can load any other Tweaker, including each other).</td></tr>
    <tr><td>Optifine</td><td><ul><li>Tweaker</li></ul></td>
        <td>Optifine is a pure Tweak-Mod without being a <i>ModSystem</i>. It hooks into the game directly but can also be co-operatively loaded by <i>ModSystems</i> such as FML and LiteLoader.</td>
    <tr><td>FML Mod Jar (container)</td><td><ul><li>Mod</li><li>Core Mod (optionally)</li></ul></td>
        <td>FML mod containers can contain both coremods and regular mods, some special treatment is given to containers which have coremods in order to grant them the power they require. Almost all features available to Tweakers are available to coremods.</td></tr>
    <tr><td>Litemod (container)</td><td><ul><li>Core Mod (Limited)</li></ul></td>
        <td>LiteLoader mods are technically limited coremods, basically because LiteLoader doesn't enforce <del>traditional gender roles</del> any particular restrictions on what litemods can do. If a litemod wants to use Mixins or Transformers then it can. LiteLoader is the <a href="http://www.urbandictionary.com/define.php?term=Aging%20Hippie%20Liberal%20Douche">Aging Hippy Liberal Douche</a> of the mod loading world.</td></tr>
    <tr><td>Bukkit Plugin (Pure)</td><td><ul><li>Plugin</li></ul></td>
        <td>Bukkit Plugins are pure plugins per the definition above.</td></tr>
    <tr><td>Bukkit Plugin (using NMS)</td><td><ul><li>Plugin</li><li>Mod</li></ul></td>
        <td>The painful truth about Bukkit plugins which use NMS is that, under the definitions above, they are basically <i>mods</i>. Any plugin which is not fully decoupled via the API puts itself into the category of Mod. The prevalence of this type of Bukkit "plugin" is due to the limitations of the API.</td></tr>
    <tr><td>Sponge Plugin (Pure)</td><td><ul><li>Plugin</li></ul></td>
        <td>Sponge Plugins are pure plugins per the definition above.</td></tr>
    <tr><td>Hybrid Mod (eg. Sponge/FML)</td><td><ul><li>Plugin</li><li>Mod</li><li>Coremod (optionally)</li></ul></td>
        <td><p>A hybrid mod leverages both a plugin component which interacts via the API, and a mod (or even coremod) in the same package. This has the disadvantages of a mod (breaks every version) but also the power of a mod (can interact with the game directly) coupled with some of the benefits of a Plugin (high-level abstract access to the game, and can also interact with other plugins as a peer).</p><p>The primary benefit of this system is that the maintenance burden is reduced when updating the mod, because any features accessed via the API are likely to be much more stable.</p><p>This type of mod can be used to implement plugins whose needs overflow the capability of the API (in the case of a plugin which needs to leverage mixins for a particular feature); but can also be used for mods which want to leverage services afforded by the API (eg. a mod which wants to provide direct support for permissions or chat channels).</p><p>Unlike NMS "plugins", a hybrid mod wears its identity on its sleeve</p></td></tr>
    <tr><td>SpongeForge</td><td><ul><li>Coremod</li></ul></td>
        <td>SpongeForge is implemented as a core mod for convenience, it exposes the Sponge API to other containers in the environment. It is not a <i>ModSystem</i> because it cannot load other containers and relies on FML for heavy lifting in that area.</td></tr>
</table>

Several things should be evident at this point:

* The *container* doesn't define the *contents*, and naming things based on the container doesn't tell the whole story, but naming things based on the contents is not in vogue. In fact it should be increasingly clear that nomenclature is only loosely coupled to reality. Given that the container in question for all these things is just a 'jar' file, ascribing a particular name to the thing doesn't tell you that much anyway: a single jar could *easily* contain a tweaker, and FML mod, a coremod, a bukkit plugin, a sponge plugin, and a litemod. What would you call it at that point?

* A bukkit "plugin" which uses NMS is more accurately a *mod*, but you'd never hear this expressed

* A sponge plugin which leverages mixin is more properly a *core mod* based on its contents, but does it matter what we call it as long as the contents/nature are understood?

* After all people were okay with CraftBukkit-NMS-using plugins still being called "plugins", so what's the deciding factor?

* Hybrid mods (hybrid plugins) are actually the most open about their contents, though the implication is generally of *mod-which-leverages-some-API-stuff* more than *plugin-which-does-extra-stuff*

Strictly speaking, the following are true:

* To use mixins in FML, you need to be a coremod. It doesn't matter if your jar contains a Sponge plugin as well, you still need to be a coremod, so most properly the container is a "hybrid mod" at this point

* To use mixins in SpongeVanilla, you need to declare your intentions in the manifest, SpongeVanilla then injects your mixins

* To have both types in a single jar is entirely possible.

So the questions are:

* Does the nomenclature matter when the name for the container clearly doesn't reflect the contents anyway?

* Is it better to just call all non-pure plugin containers something specific? eg. *Hybrid Plugin*, *Core Plugin*, etc. (Muggin? Plod?)

* Does this statement:
 
 > It is possible for plugins to use mixins, but by that point they are better classified as mods.
 
 actually say everything that is relevant in this case, given that it's so complicated and largely nobody actively gives a shit? Or is it too gross of an over-simplification which begs elaboration via a separate article?

`</walloftext>`

Tzky:
I think it's a good idea to put the above classification+definition of plugins, mods, coremods and such into a separate article on the docs. However i'm not sure where it would currently fit.

Inscrutable:
Nah, I think it'd make a magnificent *footnote*. As long as we add a flowchart.

We could add the definitions to the glossary, but they are a bit verbose for that. I'd prefer it to stay amongst best practices, unless we add a new page for it. 

... and Yes, I was of the opinion that the current single sentence answer is "an over-simplification which begs elaboration", hence my subtle request for elaboration. The issues I could see largely revolve around lessened version stability, and the fact that they ("Sponge plugin-mixins") don't necessarily have to support both implementations (eg ForgeGuard). Adding mixins does add a bit of complexity when explaining compatibility. Some good (and poor) use-cases may help guide newer developers to a sensible choice of _SpongePluginModThingembob_ architecture based on their needs.

Inscrutable:
The only sensible place I can see this information fitting is in a new page, linked from the original "Best Practices" page. I could substitute the text for something like "Mixins can be used by plugins, but be aware of the additional complexities involved. Please read the Hybrid Plugins page for more information". It'll still need heavy editing, and maybe a better Title.

The alternative is to just cut the mention of plugins from this section entirely. Which would leave readers in the dark regarding Spongedev's attitude towards hybrid plugins. My hope is to minimise confusion for both users and developers.

Mumfrey:
> Nah, I think it'd make a magnificent footnote. As long as we add a flowchart.

Here's a handy flowchart to help people figure out what type of file they are dealing with:

ST-DDT:
Maybe the flowchart is also nice to be linked in the FAQ.

Mumfrey:
> Maybe the flowchart is also nice to be linked in the FAQ.

The flowchart is a joke, so no.

PR #578, opened 2017-02-09
Fix outdated code

saladoc:
Example code compiles against API 5.0.0 without errors. The fixes are simple enough, I have not attempted to update the world generation docs as that seems to be underway already.


Spongy:

PR #594, opened 2017-04-19
Add note about load-on-startup/keep-spawn-loaded

phit:
This was changed about two months ago, but was never documented. It came up today in a discussion and I figured might as well document it, quote from blood about it http://i.imgur.com/4ih8kjK.png


phit:
As I wrote in my changes, it's in the [world config](https://docs.spongepowered.org/stable/en/server/getting-started/configuration/index.html#world-configs)..

PR #596, opened 2017-05-03
Change recommended method of EconomyService retrieval

pie-flavor:
What the title says. Storing service implementations as a member and updating it on events really seems like an anti-pattern, and there's no room for failure if you get it from the ServiceManager each time.


pie-flavor:
Whoops, screwed up commit stuff.

PR #597, opened 2017-05-03
Change recommended method of EconomyService retrieval

pie-flavor:
What the title says. Storing implementations and updating them on events really seems like an anti-pattern, and this way provides no room for error.

### DIFF
--- a/source/plugin/economy/using.rst
+++ b/source/plugin/economy/using.rst
@@ -3,9 +3,10 @@ Using the Economy API
 =====================
 
 .. javadoc-import::
-    org.spongepowered.api.event.service.ChangeServiceProviderEvent
     org.spongepowered.api.service.economy.EconomyService
     org.spongepowered.api.service.economy.account.Account
+    org.spongepowered.api.service.ServiceManager
+    org.spongepowered.api.event.service.ChangeServiceProviderEvent
 
 The Economy API unifies all economy plugins under one API. This means any plugin using the Economy API
 will be compatible with all economy plugins that implement said API. This page guides you through the steps of using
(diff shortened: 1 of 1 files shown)


source/plugin/economy/using.rst
```diff
@@ -34,18 +29,24 @@ Example: Loading the EconomyService
...
-	private EconomyService economyService;
-
-	@Listener
-	public void onChangeServiceProvider(ChangeServiceProviderEvent event) {
-		if (event.getService().equals(EconomyService.class)) {
-			economyService = (EconomyService) event.getNewProviderRegistration().getProvider();
-		}
+	Optional<EconomyService> serviceOpt = Sponge.getServiceManager().provide(EconomyService.class);
+	if (!serviceOpt.isPresent()) {
+	    // handle silly users who don't install economy plugins on their servers
```
simon816: Inappropriate comment for an unbiased documentation.

pie-flavor: It was intended as a joke - would "silly users who don't install required dependencies" be acceptable?

simon816: Why make the assumption that there must be an economy implementation?

Something like "Handle the case where an economy service is not available" would be better IMO

source/plugin/economy/using.rst
```diff
@@ -34,18 +29,24 @@ Example: Loading the EconomyService
...
-	private EconomyService economyService;
-
-	@Listener
-	public void onChangeServiceProvider(ChangeServiceProviderEvent event) {
-		if (event.getService().equals(EconomyService.class)) {
-			economyService = (EconomyService) event.getNewProviderRegistration().getProvider();
-		}
+	Optional<EconomyService> serviceOpt = Sponge.getServiceManager().provide(EconomyService.class);
+	if (!serviceOpt.isPresent()) {
+	    // handle there not being an economy implementation
```
Inscrutable: I have two quibbles here (unless they breed). Comments in code blocks don't get translated, and it's clumsy wording - how about "handle the absence of an economy implementation".

pie-flavor: How do you mean comments don't get translated? What would be an alternative?

My concern with putting actual code in there is knowing how many people would copy it verbatim.

Inscrutable: Basically, any text inside a code block gets ignored by Crowdin, so our translators never get to translate comments inside them. Hence [docs guideline 3.1](https://docs.spongepowered.org/stable/en/contributing/spongedocs.html) was born. That said, if there's no sensible alternative, then leave it where it is.

PR #606, opened 2017-06-05
Added some information about DataFormat

johnfg10:
Since i recently learned about this and though it would be useful for others i decided to put it in the docs

### DIFF
--- a/source/plugin/data/serialization.rst
+++ b/source/plugin/data/serialization.rst
@@ -171,13 +171,5 @@ For this example we will use :javadoc:`DataFormats#JSON` to translate a ItemStac
 .. code-block:: java
 
     public Optional<ItemStackSnapshot> getItemStackSnapshotFromDataContainer(DataContainer dataContainer){
-        final Optional<DataBuilder<ItemStackSnapshot>> dataBuilderOptional =            Sponge.getDataManager().getBuilder(ItemStackSnapshot.class);
-        if (dataBuilderOptional.isPresent()){
-            DataBuilder<ItemStackSnapshot> itemStackSnapshotDataBuilder = dataBuilderOptional.get();
-            Optional<DataView> dataViewOptional = dataContainer;
-            if (dataViewOptional.isPresent()){
-                return itemStackSnapshotDataBuilder.build(dataViewOptional.get());
-            }
-        }
-        return Optional.empty();
+        return Sponge.getDataManager().deserialize(ItemStackSnapshot.class, dataContainer);
     }


source/plugin/data/serialization.rst
```diff
@@ -123,3 +123,62 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
         final DataView container = translator.translate(node);
         return deserializeHealth(container);
     }
+
+
+DataFormat
+==========
+A alternative to DataTranslators is using :javadoc:`DataFormat.readFrom` or :javadoc:`DataFormat.writeTo` which allows you to convert a data container this allows you to store a data container as a hocon, json or even NBT you can of course retrive data using data format.
```
ST-DDT: You have to add your class to the javadoc-import block at the top of the page. Also I'm not sure whether you can reference methods like this.

johnfg10: ahh i see thank you and i only named the methods like that because i was guessing with the error it has been corrected :)

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,56 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
...
+        }
+
+        return outputStream;
+    }
+
+**Code Example: Serializing a PlayerContainer to json format**
+
+.. code-block:: java
+
+        public Optional<DataContainer> fromJsonFormat(InputStream inputStream){
```
Deamon5550: This is an odd example considering its just wrapping the dataformat to squash the error and returns a null instead of an optional (!).

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,56 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
         final DataView container = translator.translate(node);
         return deserializeHealth(container);
     }
+
+
+DataFormat
+==========
+A alternative to DataTranslators is using :javadoc:`DataFormat` which allows you to convert a data container this allows you to store a data container as a hocon, json or even NBT you can of course retrive data using data format.
```
Inscrutable: This sentence is a bit wonky, and far too long. How about this:
"An alternative to DataTranslators is to use :javadoc:`DataFormat`, which allows you to store a data container as a hocon, json or NBT file. You can also retrieve data using `DataFormat`."
It is also preferable to format class names etc like `this`, wherever they appear.

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,56 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
...
+
+This is very useful if your for example using a database to store information as you can then serialize any data container to for example json format
```
Inscrutable: your -> you are (or you're)
you can use e.g. instead of "for example"
it needs *at least* one comma, preferably after "information"

johnfg10:
Json is new in sponge API 6 but hocon and NBT have been in for a while basically since the beginning but uses different methods prior to API 5 
so this is for Sponge API 5-7

source/plugin/data/serialization.rst
johnfg10: done :)

source/plugin/data/serialization.rst
johnfg10: done :)

source/plugin/data/serialization.rst
johnfg10: done :)

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,56 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
         final DataView container = translator.translate(node);
         return deserializeHealth(container);
     }
+
+
+DataFormat
+==========
+An alternative to DataTranslators is to use :javadoc:DataFormat, which allows you to store a ``DataContainer`` as a hocon, json or NBT file. You can also retrieve data using ``DataFormat``.
```
Minecrell: HOCON and JSON should be written in capital letters, just like NBT

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,56 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
...
+    import java.io.IOException;
+    import java.io.OutputStream;
+    import java.io.InputStream;
+    import java.util.Optional;
+
+**Code Example: Serializing a Player to json format**
+    
+.. code-block:: java
+
+    public OutputStream toJsonFormat(Player player){
```
Minecrell: It might be better to use a specific example, like writing a `ItemStack` to a file and loading it back from it. You use an arbitrary `ByteArrayOutputStream` here which is returned as `OutputStream` so the return value of this method is essentially useless without additional casting. Additionally, how often do you actually serialize an entire `Player` on your own, and how would you load it back? The example below does only show how to read it back into a `DataContainer`.

johnfg10: i will use itemstack but i havnt written it to a file because that isnt really what i would expect people to do with it there more likly to use DataTranslators for that i was more thinking this would be used for databases etc also im not 100% sure how to go from datacontainer to the orignal object havnt been able to work that out yet

source/plugin/data/serialization.rst
johnfg10: Done :)

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,56 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
         final DataView container = translator.translate(node);
         return deserializeHealth(container);
     }
+
+
+DataFormat
+==========
+An alternative to DataTranslators is to use :javadoc:DataFormat, which allows you to store a ``DataContainer`` as a HOCON, JSON or NBT file. You can also retrieve data using ``DataFormat``.
```
dualspiral: It doesn't necessarily create files. It transforms ``DataContainer``s into HOCON, JSON or NBT _formats_, which _could_ be converted into a file, but in your example below, you don't create a file. 

That last sentence should read something like:

> You can also re-create ``DataContainer``s using these ``DataFormat``s.

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,56 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
...
+
+This is very useful if you're for example using a database to store information, as you can then serialize any ``DataContainer`` to json format.
```
dualspiral: There are a few instances in your commit still where you need to update "json" to JSON. Also, this sentence doesn't seem like it fits well - it might be better to say something like:

> For example, you can use ``DataFormat`` to create a JSON representation of a ``DataContainer``, which can then be easily stored in a database as JSON. ``DataFormat`` can then be used to rebuild the ``DataContainer`` from this JSON when required.

It's possible it could be worded better than that, but this sentence doesn't read well as it is.

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,56 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
...
+        try {
+            dataFormat.writeTo(outputStream, playerDataContainer);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        return outputStream;
+    }
+
+**Code Example: Serializing a ItemStack from json format**
```
dualspiral: That's not what this example is - it's deserializing into a `DataContainer` from `JSON`.

I think what Deamon was getting at is this full example doesn't really tell you much, all you are doing is wrapping the result in an optional, returning an empty `Optional` if it fails. It doesn't tell you much - you might be better served showing how you then could use the resultant `DataContainer`, perhaps showing how to re-create the `ItemStack` from the container? That would fit nicely with your example above - it's showing that the returned `DataContainer` could be a valid `ItemStack` - which it would be in the context of the previous example.

source/plugin/data/serialization.rst
johnfg10: fixed

source/plugin/data/serialization.rst
johnfg10: ok ive learned how to do this now and will add it to the example shortly

source/plugin/data/serialization.rst
```diff
@@ -168,22 +168,14 @@ For this example we will use :javadoc:`DataFormats#JSON` to translate a ItemStac
...
-            if (dataContainerOptional.isPresent()){
-                DataContainer dataContainer = dataContainerOptional.get();
-                final Optional<DataBuilder<ItemStack>> itemStackDataBuilder = Sponge.getDataManager().getBuilder(ItemStack.class);
-                if (itemStackDataBuilder.isPresent()){
-                    DataBuilder<ItemStack> itemStackBuilder =  itemStackDataBuilder.get();
-                    return itemStackBuilder.build(dataContainer);
-                    
-                }
+    public static Optional<ItemStackSnapshot> getItemStackSnapshotFromDataContainer(DataContainer dataContainer){
+        final Optional<DataBuilder<ItemStackSnapshot>> dataBuilderOptional =            Sponge.getDataManager().getBuilder(ItemStackSnapshot.class);
```
gabizou: All of the data manager stuff is unnecessary, all you need to do is call `dataContainer.getSerializable(ItemStackSnapshot.class, query)` or however the parameters are ordered.

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,59 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
...
+
+**Code Example: Deserializing a ItemStackSnapshot from json format**
+
+.. code-block:: java
+
+    public Optional<ItemStackSnapshot> getItemStackSnapshotFromDataContainer(DataContainer dataContainer){
+        final Optional<DataBuilder<ItemStackSnapshot>> dataBuilderOptional =            Sponge.getDataManager().getBuilder(ItemStackSnapshot.class);
+        if (dataBuilderOptional.isPresent()){
+            DataBuilder<ItemStackSnapshot> itemStackSnapshotDataBuilder = dataBuilderOptional.get();
+            Optional<DataView> dataViewOptional = dataContainer.getView(DataQuery.of("item"));
```
gabizou: Where are you getting the `item` query from?

source/plugin/data/serialization.rst
johnfg10: i tried doing that in my plugin and it didnt work correctly for me so i went with this method(as i know it works)

source/plugin/data/serialization.rst
johnfg10: fixed now sry forgot to remove it i pulled that example out of one of my plugins

source/plugin/data/serialization.rst
johnfg10: well i suppose i must of used it incorrectly the first time hmm il update the example shortly

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,51 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
...
+        try {
+            dataFormat.writeTo(outputStream, playerDataContainer);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        return outputStream;
+    }
+
+**Code Example: Deserializing a ItemStackSnapshot from json format**
```
gabizou: So I know it's been discussed before that you want to show how to use the `DataContainer` resulting from being converted from `JSON`, but I feel that it should still be shown *how* to convert it from JSON to a DataContainer, @dualspiral, I'm open for suggestions. I don't want to make it back and forth, but the wording is a little awkward since you state here that it's deserializing from JSON format, but there's no mention of JSON.

dualspiral: So, the original thought of mine is that while it doesn't matter really what we deserialise to, having an example in context is often beneficial - an `ItemStackSnapshot` is a classic example of what you might want to serialise/deserialise. Right now, this code example is not useful, because it now skips the step of _exactly_ what we want to show.

Effectively, the goal is to say "hey, if you don't want to serialise your `DataContainer` into configuration, then `DataFormat` is for you". I hate to say this, but many developers learn by example, and may be stuck on the point that `ByteArrayOutputStream`s and input streams are here, I suspect some don't know what they do and is going to detract from the actual goal. I also expect most developers will use this for databases, and just want to get a string out to save into the database. 

On the flip side, saying "here is a data container, here is JSON, here is the container again" is not going to wash either - why would a developer care about a simple data container?

So, how about, rather than providing two examples, we consider providing a sample class here that developers can use that is essentially a template on how we might want to store and retrieve a data container backed object? Something like `JsonItemStackSnapshotConverter` (or whatever you like) that contains two commented methods, `toJson` and `fromJson`, that illustrates how to get a `String` from an `ItemStackSnapshot`, and vice versa. That way, we get the context of when we might use this (item serialisation), and how this would be performed?

johnfg10: is [https://pastebin.com/VxtsLG5u](this) ok @dualspiral ?

Inscrutable:
We still have 2 pending reviews here. Can we get this cleaned up and merged soon?

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,51 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
...
+
+    import java.io.ByteArrayOutputStream;
+    import java.io.IOException;
+    import java.util.Optional;
+
+**Code Example: Serializing a ItemStackSnapshot to JSON format**
+    
+.. code-block:: java
+
+    public ByteArrayOutputStream toJsonFormat(ItemStackSnapshot itemStack){
```
Minecrell: Can you add a space before `{`

source/plugin/data/serialization.rst
```diff
@@ -123,3 +125,51 @@ from persistent files using the :doc:`Configurate Library <../configuration/inde
...
+        DataContainer playerDataContainer = itemStack.toContainer();
+        DataFormat dataFormat = DataFormats.JSON;
```
Minecrell: I don't think it makes any sense to store this in a variable first, just access it directly using `DataFormats.JSON.writeTo` below.

source/plugin/data/serialization.rst
Minecrell: I also think it would be easier to understand if you don't return the `ByteArrayOutputStream`, but rather use a specific example like writing a `ItemStackSnapshot` to a file. I don't think anyone would use the method like this.

johnfg10:
ive had to open a new pull request as i apparently deleted the fork as some point

source/plugin/data/serialization.rst
johnfg10: done in the new pr

source/plugin/data/serialization.rst
johnfg10: done in the new pr

Issue #364, opened 2015-11-05
Tutorial on SpongeForge event implementation?


JBYoshi:
Moved from SpongePowered/SpongeForge#379.

Inscrutable:
Is there more to the [Events page](https://docs.spongepowered.org/en/plugin/events.html?highlight=events) that you need us to add?

saladoc:
@JBYoshi please edit your description to include an actual statement on what you request / ask of the docs. That will be easier for us than to first read the whole linked issue.

JBYoshi:
@Inscrutable This isn't for plugin using events. This is for implementing events (see the updated description).

@saladoc Done.

ST-DDT:
The link in the description is no longer working.

JBYoshi:
Actually, most of this system was rewritten in https://github.com/SpongePowered/SpongeForge/commit/20734562847bcb60cade1611fbab74bc75b47978. I replaced the link, though, and pointed it to one of the few remnants.

Issue #231, opened 2015-05-22
Information for Mod developers to add compatibility?


ZephireNZ:
So Sponge is obviously being designed with mod compatibility in mind. We should create guides for mod developers to add APIs to access their mod features, data, etc through Sponge.

Guides for:
- DataManipulators. this would be a major one - the ability for mods to add manipulators for their custom entities, items and blocks, or adding support for existing manipulators. Should also talk about ideals (abstract similar data together, not all of an entities metadata, for example)
- BlockType/ItemType/etc. Registering types in the registry, as well as builders
- Inventories. Another big section that mods change, and how they should support our Inventory Query system.

There's probably more that I've forgotten as well!

ryantheleach:
Worldgen,

World modify on behalf of entities / players / machines

Inventory Transfer

Attack / damage sources / tracking.

properties, 

pvp control

projectile and machine owners

entity owners where applicable.

Off-world storage.

BlackGoat:
From the Best Practices doc (WIP), there is this (may be outdated and not definitive):
*How should devs approach using Forge or NMS classes?*
Mods that add to the SpongeAPI using internals will have to specifically write an API that isn’t relying on underlying Minecraft code to be usable by Sponge plugins. However, plugins can be created that load separate “compatibility” modules to interact with the underlying implementation (of which there are two major implementations, Sponge for Forge, and SpongeVanilla).
*Are there any specific areas that will be left out of the API, that will necessitate invoking Forge/NMS?*
Anything to do with intercepting packets, adding NPC’s, introducing custom items/blocks/entities/etc.

BlackGoat:
A new section on Mod interactions with Sponge API would be a useful addition. The Plugins section would be appropriate, next to Basic and Advanced. Presumably this will only cover Forge mods?
Whomsoever wants to write something helpful to get this started, please accept this Green Light.

PR #621, opened 2017-08-26
Simple Forge Permission

MoeBoy76:
- added section to Permissions doc on how to create a Sponge permission
by using Forge on its own

### DIFF
--- a/source/plugin/permissions.rst
+++ b/source/plugin/permissions.rst
@@ -405,4 +405,4 @@ It is recommended to follow the permission structure as described above. The per
 
 .. note::
     
-    The srg name for this method is ``func_70003_b``
+    The SRG name for this method is ``func_70003_b``.
--- a/source/plugin/permissions.rst
+++ b/source/plugin/permissions.rst
@@ -372,11 +372,15 @@ The ``ContextCalculator`` can be registered via:
     permissionService.registerContextCalculator(contextCalculator);
 
 For Forge Mods
-========
+==============
 
-If you are the author of a Forge mod and are not using the new Forge PermissionsAPI but are doing OP checks then you are already on the right path for Sponge to pick up permissions.
+If you are the author of a Forge mod and are not using the new Forge PermissionsAPI but are doing OP checks, then you are already
+on the right path for Sponge to pick up permissions.
 
-The simplest way to create a Sponge permission in a Forge mod without soft-depending on SpongeAPI is to use the method provided by Vanilla Minecraft code in ``ICommandSender``, namely ``ICommandSender.canCommandSenderUseCommand(int permLevel, String commandName)``. The String passed into that method has no use at all in a Vanilla Forge environment but when Sponge Forge is added it automatically takes that String and converts it into a working permission.
+The simplest way to create a Sponge permission in a Forge mod without soft-depending on SpongeAPI is to use the method provided by
+Vanilla Minecraft code in ``ICommandSender``, namely ``ICommandSender.canCommandSenderUseCommand(int permLevel, String commandName)``.
+The String passed into that method has no use at all in a Vanilla Forge environment, but when SpongeForge is added it automatically
+takes that String and converts it into a working permission.
 
 Example
 ~~~~~~~
@@ -394,7 +398,9 @@ Example
         }
     }
 
-As you can see, we simply check for the OP level and pass in a random String we want to use as a permission when Sponge is used. When Forge is used by itself the player simply requires the OP level, passing a value of 0 would allow all users to interact with the block but when SpongeForge is added they require the permission node of ``examplemod.awesomeblock.interact``.
+As you can see, we simply check for the OP level and pass in an arbitrary String we want to use as a permission when Sponge is used.
+When Forge is used by itself the player simply requires the OP level, so passing a value of 0 would allow all users to interact with
+the block, but when SpongeForge is added they require the permission node of ``examplemod.awesomeblock.interact``.
 
 .. note::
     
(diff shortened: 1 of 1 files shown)


source/plugin/permissions.rst
```diff
@@ -370,3 +370,32 @@ The ``ContextCalculator`` can be registered via:
 .. code-block:: java
     
     permissionService.registerContextCalculator(contextCalculator);
+
+For Forge Mods
+========
```
Inscrutable: Travis failed the build because this underline was too short (they have to match the length of the title line).

source/plugin/permissions.rst
```diff
@@ -370,3 +370,32 @@ The ``ContextCalculator`` can be registered via:
...
+
+If you are the author of a Forge mod and are not using the new Forge PermissionsAPI but are doing OP checks then you are already on the right path for Sponge to pick up permissions.
+
+The simplest way to create a Sponge permission in a Forge mod without soft-depending on SpongeAPI is to use the method provided by Vanilla Minecraft code in ``ICommandSender``, namely ``ICommandSender.canCommandSenderUseCommand(int permLevel, String commandName)``. The String passed into that method has no use at all in a Vanilla Forge environment but when Sponge Forge is added it automatically takes that String and converts it into a working permission.
```
Inscrutable: SpongeForge should be one word. Also, line lengths should be shorter (120 characters or less) to meet our usual guidelines.

source/plugin/permissions.rst
```diff
@@ -370,3 +370,32 @@ The ``ContextCalculator`` can be registered via:
...
+        public boolean onBlockActivated(World world, BlockPos pos, IBlockState state, EntityPlayer player, EnumFacing side, float hitX, float hitY, float hitZ) {
+            if(player.canCommandSenderUseCommand(4, "examplemod.awesomeblock.interact")) {
+                //Do cool stuff
+                return true;
+            }
+        return false;
+        }
+    }
+
+As you can see, we simply check for the OP level and pass in a random String we want to use as a permission when Sponge is used. When Forge is used by itself the player simply requires the OP level, passing a value of 0 would allow all users to interact with the block but when SpongeForge is added they require the permission node of ``examplemod.awesomeblock.interact``.
```
Inscrutable: I think you mean "arbitrary", not "random".

source/plugin/permissions.rst
```diff
@@ -370,3 +370,32 @@ The ``ContextCalculator`` can be registered via:
...
+
+.. note::
+    
+    The srg name for this method is ``func_70003_b``
```
Inscrutable: I'm not sure this will be much use here, as it depends on the MCP mappings, which aren't referenced much in these docs so far, but it could be important for Forge mod devs. I am seeking opinions.

MoeBoy76: The problem with not having it is if they change the name of the method (it used to be hasPermission), plus this is a section about Forge development itself

ST-DDT: I agree. The name is not dependend on Sponge. Users that already use this kind of OP check already use it. Experienced users will propably use the 'fixed' name anyway. Inexperienced users are unlikely to try to be compatible with Sponges PermSystem. And if they try its probably easy enough to just search for it or even guess it from the parameters.

ST-DDT: IMO Name changes to the remapped names are less likely than the raw names.

ST-DDT:
> I also want some opinions on where this belongs - as an extension to the existing page (as now), or a new page in a perms hierarchy, or new section on forge modding with Sponge?

Might be good in the Forge modding section with a link from and to the perm page. Otherwise it will be hard to find it.

source/plugin/permissions.rst
MoeBoy76: Hence the usage, the searge name never changes but the MCP mapping name does

source/plugin/permissions.rst
MoeBoy76: These are exactly the kinds of things i need reviewed, sometimes i use too high level language and other times i don't

Inscrutable:
@ST-DDT The section "Implementation-dependent Plugins" could be expanded to contain Forge-specific stuff. Otherwise we'd need a new page and/or folder. There are a couple of outstanding issues on interacting with Forge (#231, #364) that also need a place to live.

source/plugin/permissions.rst
Inscrutable: If it's fixed for a given SpongeAPI version (which all conveniently happen to belong to different Minecraft versions), then it should be OK. If this PR were to be cherry-picked to earlier API versions, we'd need separate updates.

source/plugin/permissions.rst
MoeBoy76: I had done a quick search for the word Forge earlier and noticed that Sponge Forge came up in the majority of the results, I thought it was odd and I will note it in the OCD PR

Inscrutable:
OK, I've cleaned up most of the problems here. @ST-DDT may still want to add something about the permission string format, and we could use info on which version the srg name is for (so we can swap it out for API 5/6/7 docs).

source/plugin/permissions.rst
```diff
@@ -370,3 +370,39 @@ The ``ContextCalculator`` can be registered via:
...
+Vanilla Minecraft code in ``ICommandSender``, namely ``ICommandSender.canCommandSenderUseCommand(int permLevel, String commandName)``.
+The String passed into that method has no use at all in a Vanilla Forge environment, but when SpongeForge is added it automatically
+takes that String and converts it into a working permission.
+
+Example
+~~~~~~~
+
+.. code-block:: java
+
+    public Block AwesomeBlock extends Block {
```
Minecrell: This line doesn't make any sense. I guess you mean `public class AwesomeBlock`

PR #622, opened 2017-09-02
Improve database connection page

phroa:
- try-with-resources to autoclose the connection and statement
- say something about getting the connection urls
- javadoc DataSource


phroa:
k

Inscrutable:
... and now I realise it was targeting "master" not "stable". AAARGH

PR #628, opened 2017-09-24
modifiers.rst public instead of private class

vorburger:
Thre is no "private class" in Java... ;-)


ST-DDT:
Well for such small (and uncontested) changes use the [OCD issue](https://github.com/SpongePowered/SpongeDocs/issues/437) next time (or commit it straight to stable).

kashike:
For what it's worth, it *is* possible to have `private` classes in Java - just not at the root level, and not when you need to access it externally, such as in this case.

PR #629, opened 2017-10-11
Correct list-users ore route

Dannyps:
Solves https://github.com/SpongePowered/SpongeDocs/issues/437#issuecomment-324971059

### DIFF
--- a/source/ore/routes/list-users.rst
+++ b/source/ore/routes/list-users.rst
@@ -11,9 +11,9 @@ Returns a list of users based on given criteria.
 +--------+-----------+-------------------------------------------------------------+
 | Name   | Data Type | Description                                                 |
 +========+===========+=============================================================+
-| limit  | Integer   | Limits the amount of projects returned (max / default: 25). |
+| limit  | Integer   | Limits the amount of users returned (max / default: 25).    |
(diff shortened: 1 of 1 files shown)


Inscrutable:
I'm happy to share the love, and let this get merged, as it helps whittle down the OCD list a bit. It is starting to look daunting again, and some of the changes needed are quite extensive. So if those end up bundled into separate PR's too I'm happy, as long as everyone knows who's doing what.

Dannyps:
I submitted this PR because it's been more than a month since I wrote the comment on the OCD list, and the issue was still there. Because the list is growing so big, perhaps it would be better if we all PR'ed our corrections to `minor-fixes-ocd`.

ST-DDT:
@Dannyps There are a lot of other issues on the OCD list, if you are willing you could also create a bigger PR for the other ones that don't require further discussion (most stuff after the SpongeAPI vs Sponge API comments).
If you propose the changes, I'll review them.
If you don't have time, I will try to do it [soon](https://www.urbandictionary.com/define.php?term=soon#4387130). 
If you are willing to do it please leave a comment in the OCD, otherwise leave a message here, so I can (prepare to) start early

PR #631, opened 2017-10-22
Minor fixes ocd

Dannyps:
Addresses several issues defined in #437.

### DIFF
--- a/source/plugin/blocks/accessing.rst
+++ b/source/plugin/blocks/accessing.rst
@@ -105,7 +105,7 @@ Block States
 ~~~~~~~~~~~~
 
 A :javadoc:`BlockState` contains a :javadoc:`BlockType`,  any ``DataManipulator``\ s and properties that are applied to
-the block, and any :javadoc:`BlockTrait`\ s for a block. It stores all immutable value's for a particular block. One
+the block, and any :javadoc:`BlockTrait`\ s for a block. It stores all immutable values for a particular block. One
 use of this is getting an :javadoc:`ImmutableDataManipulator`, as shown below:
 
 .. code-block:: java
--- a/source/plugin/blocks/concepts.rst
+++ b/source/plugin/blocks/concepts.rst
@@ -38,7 +38,7 @@ not utilizing their metadata at all. On the other hand, logs do use their metada
 properties exceeds 16 possible values (think tree type and direction), logs must be split over two base types.
 
 In the future, there will only be one 16 bit number (65536 possible combinations of base type + properties). Blocks will
-be assigned an ID automatically and this assignment will be stored in the world save file. This is illustrated below::
+be assigned an ID automatically and this assignment will be stored in the world save file. This is illustrated below:
 
     0 => minecraft:dirt[snowy=false,variant=default]
     1 => minecraft:dirt[snowy=true,variant=default]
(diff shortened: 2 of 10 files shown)


Dannyps:
I solved all the issues to which I reacted with a  ❤️  :

Inscrutable:
One question @Dannyps : it seems to be targeting ``SpongePowered:minor-fixes-ocd`` not ``SpongePowered:stable``. Is this intentional?

ImMorpheus:
I've added dead link/redirect to the OCD list.

Dannyps:
@Inscrutable It wasn't intentional. I forked `SpongePowered:minor-fixes-ocd` to prepare the PR, that's why I created it targeting that specific branch. Should I change to `SpongePowered:stable`? I believe it's possible.

Inscrutable:
Hmmm. I'd prefer it went to the stable branch, unless you and @ST-DDT are planning an omnibus ocd-PR.

ST-DDT:
Please target `stable`.

Dannyps:
@Inscrutable, @ST-DDT, done!

ST-DDT:
By changing the target you also took all the other commits from the other branch with you.
The easiest way to do this is to create a backup of your branch (with a different name), reset it to latest stable and then cherry pick your commits over.

Dannyps:
I'm afraid I'm not very comfortable with PRs 😟
Would that require submitting a new Pull Request?

ST-DDT:
No. It does not all you have to do is force pushing the fixed branch.
If you are not comfortable with it i can help you with it in roughly a week or two.

Dannyps:
@ST-DDT I'd appreciate that, thank you.

ST-DDT:
I fixed the commit history. If you need to make more adjustments make sure to use a clean/force pull before adding them. Old rev; https://github.com/Dannyps/SpongeDocs/commit/ee301cc01a99f2d2267961d6c6afd102767321d8

----------------------------------

Also I have seen that you reverted a commit because it causes the build to fail.
You can fix it be adding it to the imports at the top of the file like this:

````rst
.. javadoc-import::
	org.spongepowered.api.text.Text
	org.spongepowered.api.text.Text.Builder
	org.spongepowered.api.text.action.TextAction
````

If you need to attach a plural to those references you can do so using a backslash.

````rst
and :javadoc:`TextAction`\s.
````

-------------------------------

Anything else you would like to add/change?

Dannyps:
Thank you, @ST-DDT! About that reverted commit, I'll correct it.

As for other stuff, I can't do much these days because of college. It'd probably be better if this could be merged when I done with that last fix, and we could have another PR like this somewhere in the future, to address new issues.

Inscrutable:
Ummm, is this ready yet?

Dannyps:
Nope, I'm still to fix that reverted commit. I believe I'll have that done in about 24h.

Dannyps:
Replying to https://github.com/SpongePowered/SpongeDocs/pull/631#issuecomment-343765271
, and about fixing that plural, it looks like `:javadoc:*` doesn't support the `\s` terminology.

Examples: https://docs.spongepowered.org/stable/en/plugin/permissions.html#subject
(notice the `s` in `Players` is not part of the anchor)

I'll put it this way, but we should try to fix this in the future.

Dannyps:
Should I open an issue addressing the the `\s` behaviour of `:javadoc:*`?

ryantheleach:
The plural thing is a PITA, It's good as is IMO, it's only going to get twisted for the translations anyway.

PR #643, opened 2017-12-05
Mixin does allow lambdas and method references.

JBYoshi:
Mixin originally did not support the Java <s>9</s> 8 lambdas and method references, but it was later updated to include them. This fixes the docs to indicate that.

### DIFF
--- a/source/contributing/implementation/mixins.rst
+++ b/source/contributing/implementation/mixins.rst
@@ -30,8 +30,8 @@ repository itself, since almost everything is already documented.
 Mixins and Inner Classes
 ========================
 
-While you can use lambdas, along with anonymous and inner classes, inside mixins, you cannot use an anonymous/inner class within another
-anonymous/inner class that is also inside a mixin, unless one of the inner classes is ``static``.
+While you can use lambdas, anonymous and inner classes inside mixins, you cannot use an anonymous/inner class within
+another anonymous/inner class that is also inside a mixin, unless one of the inner classes is ``static``.
 
 This means expressions like the following will cause mixins to fail horribly and bring death and destruction upon all
 that attempt to use Sponge.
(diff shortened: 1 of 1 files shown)


Spongy:

ryantheleach:
Java 8?

JBYoshi:
🤦‍♂️

ST-DDT (approved):

Mumfrey:
You can use anonymous and non-static inner classes as well now, but only to a nested depth of *1*. Eg. you can have a mixin class containing an inner class or an anon inner class, but you can't have a mixin containing an inner class which *itself* contains an anonymous or inner class. See [this](https://github.com/SpongePowered/Mixin/wiki/Flippin'-Mixins,-how-do-they-work%3F#3311-inner-class-generator) for details of how this is handled and why the depth is limited.

Inscrutable:
Is it worth incorporating Mumfrey's comments here, or shall I merge this as-is, @JBYoshi ?

JBYoshi:
@Inscrutable Just updated it.

source/contributing/implementation/mixins.rst
```diff
@@ -21,49 +21,116 @@ the `SpongeCommon repository <https://github.com/SpongePowered/SpongeCommon/tree
...
+.. note::
+
+  The Mixin project will be servicing a number of other projects in addition to Sponge itself. Therefore Mixin has its'
+  own documentation together with the repository.
+
+
+Mixins and Inner Classes
+========================
+
+While you can use lambdas, along with anonymous and inner classes, inside mixins, you cannot use an anonymous/inner class within another
```
Inscrutable: Too many commas. May I suggest the following edit:
"While you can use lambdas, anonymous and inner classes inside mixins, you cannot use an anonymous/inner class within another ..."

Inscrutable (approved):
Correct the comments about commas and we can conclude this chapter.

PR #645, opened 2018-01-06
Bump Docs to 7.0.0

Inscrutable:
Also extend copyright to 2018


Spongy:

source/plugin/debugging.rst
```diff
@@ -54,7 +54,7 @@ adding the plugin project to Sponge's classpath will rid you of the necessity to
 file into your server directory after every change.
 
 First you need to ensure that you have your Run/Debug Configuration(s) set appropriately, as shown in the Sponge
-`README.md <https://github.com/SpongePowered/SpongeForge/blob/stable-6/README.md#Running>`_
+`README.md <https://github.com/SpongePowered/SpongeForge/blob/stable-7/README.md#Running>`_
```
ST-DDT: Note to self: Check whether this needs to be `#running`.

Inscrutable:
OK, I think that's all we need to at least bump the version to API 7 and let us get started. There will be a lot of files that need tinkering (especially in *Plugins*). I haven't tweaked the descriptions for the ``global.conf`` settings either, a few defaults changed and asynch lighting got threads. Let me know, or add a commit if I missed anything important.

source/plugin/debugging.rst
ST-DDT: Yeah, the hook needs to be lowercased.

Issue #619, opened 2017-08-15
Addition of '/sponge:help' permission


Grinch:
As of [this commit](https://github.com/SpongePowered/SpongeCommon/compare/2af6807b35d6...4e82a4ec9965) `/sponge:help` has the permission `sponge.commands.help` tied to it.

Inscrutable:
Presuming this targets bleeding (API 7), will it get cherry-picked into older API versions?

Grinch:
The only version I targeted for this is 7. It could be pulled into older versions but that isn't something I personally plan to PR.

PR #646, opened 2018-01-06
API7 permission nodes

phit:
closes https://github.com/SpongePowered/SpongeDocs/issues/619


Inscrutable:
Might as well tack on the new perm for the force-gamemode override too.

phit:
oh yeah, forgot about that will add in a bit

phit:
ready for review

PR #647, opened 2018-01-06
Add docs on the new child command option in CommandSpec.Builder

dualspiral:
This is for API 7, don't merge this for API 6 docs before it gets split off!

In API 7, a new option for child commands was added to allow developers choose what would happen if a child command couldn't parse its arguments AND there was a fallback/parent executor - `CommandSpec.Builder#childArgumentParseExceptionFallback`. I've added information on how child command parsing works in API 7 - review on the clarity, headings, wording and the en_US spelling is appreciated.


Spongy:

PR #654, opened 2018-01-30
Fix and update plugin/blocks pages

### LINKED ISSUE #602: Cause accepting plugin instance or plugin container?
I probably need someone to verify I didn't make a mistake, but I believe all the plugin instances on this page, need to be replaced with plugin containers instead.

https://docs.spongepowered.org/stable/en/plugin/blocks/modifying.html?highlight=modifying%20blocks

MinusKube:
Since the `plugin/block` pages were outdated since new 7.0.0 API, I've fixed this to make the code and the text match the new version.

This fixes #602.

### DIFF
--- a/source/plugin/blocks/accessing.rst
+++ b/source/plugin/blocks/accessing.rst
@@ -154,7 +154,7 @@ get the blast resistance of a block and checking if it is greater than or equal 
 This will get the blast resistance of our block and compare it to a new :javadoc:`DoubleProperty`, as
 :javadoc:`BlastResistanceProperty` inherits from ``DoubleProperty``. The method will then return if the blast
 resistance of our block is greater than one, the value in placed ``matches()``. If we wanted to see if it was less than
-two, we would replace it with ``Property.Operator.LESS``.
+two, we would replace the property operator in the ``DoubleProperty`` ctor  with ``Property.Operator.LESS``.
 
 If we were comparing two pre-existing properties, it will take the ``Operator`` of our first value, the one we are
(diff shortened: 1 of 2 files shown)


phit (approved):

source/plugin/blocks/accessing.rst
```diff
@@ -153,7 +154,7 @@ get the blast resistance of a block and checking if it is greater than or equal
 This will get the blast resistance of our block and compare it to a new :javadoc:`DoubleProperty`, as
 :javadoc:`BlastResistanceProperty` inherits from ``DoubleProperty``. The method will then return if the blast
 resistance of our block is greater than one, the value in placed ``matches()``. If we wanted to see if it was less than
-two, we would replace it with ``lessThan()``.
+two, we would replace it with ``Property.Operator.LESS``.
```
dualspiral: What is "it"? It's not necessarily clear - specify that you can replace the property operator in the `DoubleProperty` ctor with this.

source/plugin/blocks/modifying.rst
```diff
@@ -17,29 +17,27 @@ Modifying Blocks
 Changing a Blocks Type
```
dualspiral: I know you didn't change this, but it's otherwise going to irritate me - can you add an apostrophe in "Block's" please - the block owns the type! :)

dualspiral (commented):
Looks fine, just checked the code and all matches up, just a minor readability quibble and a minor gramatical quibble and I think we're good to go.

Issue #640, opened 2017-11-29
Outdated imports in docs


ST-DDT:
I checked the docs for imports that are no longer working with v6.0 and found these:

EDIT: (Removed outdated list)

The test tool to collect all used imports:

````java
public class ImportTest {

    private static Set<String> imports = new TreeSet<>();
    private static int i = 0;

    public static void main(String[] args) {
        File root = new File("../SpongeDocs/source");
        fileStream(root).forEach(ImportTest::test);
        imports.forEach(s -> System.out.println(s + " a" + i++ + " = null;"));
    }

    static Stream<File> fileStream(File file) {
        if (file.isDirectory()) {
            return Arrays.stream(file.listFiles()).flatMap(JavadocRefTest::fileStream);
        } else if (file.isFile() && file.getName().endsWith(".rst")) {
            return Stream.of(file);
        } else {
            return Stream.empty();
        }
    }

    static void test(File file) {
        try {
            Files.lines(file.toPath())
                    .dropWhile(Predicate.isEqual(".. javadoc-import::").negate())
                    .skip(1)
                    .takeWhile(Predicate.isEqual("").negate())
                    .map(String::trim)
                    .forEach(imports::add);
        } catch (IOException e) {}
    }
}
````

ST-DDT:
And these imports are no longer used:

EDIT: (Removed outdated list)

Collected by this tool:

````java
public class JavadocRefTest {

    private static Set<String> unusedImports = new TreeSet<>();
    private static Set<String> missingImports = new TreeSet<>();

    public static void main(String[] args) {
        File root = new File("../SpongeDocs/source");
        fileStream(root).forEach(JavadocRefTest::test);
        unusedImports.forEach(s -> System.out.println(s.substring(root.getPath().length() + 1)));
        missingImports.forEach(s -> System.out.println(s.substring(root.getPath().length() + 1)));
    }

    static Stream<File> fileStream(File file) {
        if (file.isDirectory()) {
            return Arrays.stream(file.listFiles()).flatMap(JavadocRefTest::fileStream);
        } else if (file.isFile() && file.getName().endsWith(".rst")) {
            return Stream.of(file);
        } else {
            return Stream.empty();
        }
    }

    private static final Pattern PATTERN_CLASSNAME = Pattern.compile("([A-Z].+)");
    private static final Pattern PATTERN_JAVADOC = Pattern
[...]

Issue #651, opened 2018-01-24
Outdated references on Event Causes page.


LazoYoung:
[This part](https://docs.spongepowered.org/stable/en/plugin/event/causes.html#named-causes) describes how to retrieve named objects in cause chain.

```java
@Listener
public void onGrow(ChangeBlockEvent.Grow event) {
    Optional<Player> notifier = event.getCause().get(NamedCause.NOTIFIER, Player.class);
}
```

The sample code refers to `Cause#get(String, Class)` and `NamedCause`  which don't exist in the current javadoc for SpongeAPI. I guess those were replaced by EventContext since [the refactor on event causes](https://github.com/SpongePowered/SpongeAPI/issues/1184)

ST-DDT:
Probably related : https://github.com/SpongePowered/SpongeDocs/issues/640

LazoYoung:
Document also needs to provide the usage for EventContext, replacing the old NamedCause examples.

PR #656, opened 2018-02-03
Update the event cause page for API 7

dualspiral:
One of the biggest changes in API 7 was the introduction of the `CauseStackManager`. There has been some confusion over it and this page ia currently describing API 6 behaviour, so I thought I'd update it in the hope that it would then be useful for those confused about how to get the current cause.

It could do with a look at it by a developer familar with the cause stack manager to make sure I've not made things too complex.

### DIFF
--- a/source/plugin/event/causes.rst
+++ b/source/plugin/event/causes.rst
@@ -162,6 +162,17 @@ Before adding your own causes, you should push a cause stack frame to the manage
 as a saved state, when you are done with your causes, the removal of the frame returns the manager to
 its original state.
 
+.. tip::
+
+    Adding a frame to the CauseStackManager does not remove what is already in the manager, so anything
+    that is in the cause stack and contexts before a stack frame is added will be there afterwards. You
+    can verify this by calling ``Sponge.getCauseStackManager().getCurrentCause()`` before and after the
+    frame is pushed.
+
+    For example, if the cause stack contains a ``PluginContainer`` and a ``CommandSource`` when a frame
+    is pushed, they will remain on the stack and will form part of the ``Cause`` if one is obtained from
+    the frame.
+
 For example, if you were to fire an event that was simulating another player in a sudo like command,
 you may want to add your acting player in the cause and the game profile of the player that you are
 simulating in the context (as the simulated player is not directly responsible for the event being fired.)
--- a/source/plugin/event/causes.rst
+++ b/source/plugin/event/causes.rst
@@ -17,9 +17,11 @@ Event Causes
     org.spongepowered.api.event.cause.EventContextKeys
     org.spongepowered.api.event.entity.DamageEntityEvent
     org.spongepowered.api.plugin.PluginContainer
+    org.spongepowered.api.profile.GameProfile
     java.lang.Class
     java.lang.Object
     java.lang.String
+    java.util.Stack
 
 Events are great for attaching additional logic to game actions, but they have the drawback of providing next to no
 context as to what has **caused** that event to occur. The :javadoc:`Cause` object allows providing and receiving
@@ -141,7 +143,7 @@ thread or async.
 Using the CauseStackManager
 ===========================
 
-.. note::
+.. warning::
 
     The ``CauseStackManager`` only works on the main server thread. If you call it from a
     different thread, an ``IllegalStateException`` will be thrown. Ensure you are on the main
@@ -165,6 +167,11 @@ you may want to add your acting player in the cause and the game profile of the 
 simulating in the context (as the simulated player is not directly responsible for the event being fired.)
 
 **Creating a custom Cause with the CauseStackManager**
+
+In this example, the variables would be populated, the cause would contain the ``playerToSimulate`` as
+the root cause, the ``sourceRunningSudo`` as the second object in the cause and the :javadoc:`GameProfile`
+as the :javadoc:`EventContextKeys#SIMULATED_PLAYER` context, in addition to anything already in the
+``CauseStackManager``. Your event code would be at the bottom of the method.
 
 .. code-block:: java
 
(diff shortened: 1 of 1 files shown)


Spongy:

dualspiral:
Fixes #640, #651

source/plugin/event/causes.rst
```diff
@@ -69,55 +97,137 @@ object matching the provided type.
...
+    // In your code these would be populated
+    CommandSource sourceRunningSudo;
+    Player playerToSimulate;
+    PluginContainer plugin;
+
+    Cause cause = Cause.builder()
+      .append(playerToSimulate) // This would be root
+      .append(sourceRunningSudo)
+      .append(plugin)
+      .build(EventContext.builder().add(EventContextKeys.PLAYER_SIMULATED, playerToSimulate.getProfile()).build());
```
ST-DDT: IMO this should be formatted differently.
Either move it to its own variable or apply the same line breaks as above.

source/plugin/event/causes.rst
```diff
@@ -69,55 +97,137 @@ object matching the provided type.
...
+
+For example, if you were to fire an event that was simulating another player in a sudo like command,
+you may want to add your acting player in the cause and the game profile of the player that you are
+simulating in the context (as the simulated player is not directly responsible for the event being fired.)
+
+**Creating a custom Cause with the CauseStackManager**
+
+.. code-block:: java
+
+    // In your code these would be populated
```
ST-DDT: Maybe move these outside of the code blocks to make it tranlateable?

source/plugin/event/causes.rst
```diff
@@ -69,55 +97,137 @@ object matching the provided type.
...
+      // the root player is a simulated one and may not have invoked the event.
+      frame.addContext(EventContextKeys.SIMULATED_PLAYER, playerToSimulate.getProfile());
+
+      // Add your event code here.
+      // You can get the current cause at any time using the getCurrentCause method
+      // (which is available on the fram as well as the CauseStackManager)
+      Cause cause = frame.getCurrentCause();
+    }
+
+Note that the last item you push to the cause stack will be the root of the ``Cause``.
```
gabizou: May want to link to the documentation of the Stack data structure to explain that it's a LIFO collection, as some developers may get confused about why ordering matters.

source/plugin/event/causes.rst
dualspiral: It might be difficult because ideally the comments are attached to a line of code, but I'll do my best to add a summary of what I'm trying to convey. I'm inclined to leave the lines in regardless of whether I do that to maximise clarity for those who can read the comments.

Zidane:
Would it be worth reinforcing the idea that creating a frame gives you objects in the cause from the global context? I mean you do mention their container already being in the cause but developers may be slightly taken back by seeing more than their container there.

Food for thought.

dualspiral:
Sure. The more we have in the docs about this the better if you ask me.

source/plugin/event/causes.rst
```diff
@@ -25,8 +36,27 @@ creating a multitude of subevents for the different source conditions this infor
...
+  order, they are all equally important.
+
+As an example, if a sheep owned by a player eats some grass, the most direct cause of this is the sheep. The
+player would be in the ``EventContext`` as the :javadoc:`EventContextKeys#OWNER`, giving event consumers
+that additional information about how the event has come about, but would not necessarily be within the
+direct cause itself.
+
+Another example that you may need to watch out for is if you simulate a player. The simulated player may not be
+in the direct cause, as the player being simulated may not have been involved in the action, however, the player's
+be in the ``EventContext`` under the :javadoc:`EventContextKeys#PLAYER_SIMULATED`
```
Inscrutable: The syntax here is funny peculiar. "the player's be in the" sounds like Pirate to me ...

source/plugin/event/causes.rst
```diff
@@ -25,8 +36,27 @@ creating a multitude of subevents for the different source conditions this infor
...
+
+Retrieving objects from the direct cause
```
Inscrutable: Please use Title Case for Titles, as recommended in the docs guidelines. (Yes, it was bad before your edit too).

source/plugin/event/causes.rst
```diff
@@ -69,55 +99,164 @@ object matching the provided type.
...
+    different thread, an ``IllegalStateException`` will be thrown. Ensure you are on the main
+    server thread **before** calling methods on the ``CauseStackManager``.
+
+If you are creating your event on the main thread, then use the :javadoc:`CauseStackManager`, which can
+be found at :javadoc:`Sponge#getCauseStackManager()`. The ``CauseStackManager`` tracks the potential
+causes of events as the game runs, allowing for easy retrieval of the current ``Cause`` without effort.
+To see the current cause, call :javadoc:`CauseStackManager#getCurrentCause()`. You may notice that your
+plugin's :javadoc:`PluginContainer` is already in the returned ``Cause``, as plugins are one of the
+objects tracked by the manager. Using the ``CauseStackManager`` for creating causes removes the
+need for boilerplate-like code where you supply objects like your plugin container, you can concentrate
```
Inscrutable: This sentence is incomplete, it needs a "so". i.e. "... container, *so* you can concentrate ..."

source/plugin/event/causes.rst
```diff
@@ -69,55 +99,164 @@ object matching the provided type.
...
+      frame.pushCause(playerToSimulate);
+
+      // This would contain the GameProfile of the player to simulate, they are not a direct
+      // cause, but provide additional information to the event listeners so that they know
+      // the root player is a simulated one and may not have invoked the event.
+      frame.addContext(EventContextKeys.SIMULATED_PLAYER, playerToSimulate.getProfile());
+
+      // Add your event code here.
+      // You can get the current cause at any time using the getCurrentCause method
+      // (which is available on the fram as well as the CauseStackManager)
```
Inscrutable: I will concur with the observation that you put a lot of comments in the code blocks (compared to existing docs). These cannot be translated by _Crowdin_ (it sensibly ignores code blocks), so it does limit their reach somewhat. We prefer comments were left *outside* the code blocks wherever possible, for a host of reasons. There _are_ numerous good reasons to have comments in there; are you sure this is one of those cases?

dualspiral:
I've taken the comments out and addressed other comments. I've not done anything about the commas yet, but please feel free to add them as you feel necessary, I won't be upset if you tinker with the PR!

Inscrutable:
Anything else is optional; it's still readable. I didn't expect you to purge the code comments _entirely_, but it does make it shorter. Unless anyone else has any quibbles I'll merge this shortly.

PR #670, opened 2018-03-06
Document Ore deployment API

Katrix:
Documents the Ore deployment API, including the two recent pull requests to add more options to it.

### DIFF
--- a/source/ore/routes/deploy-version.rst
+++ b/source/ore/routes/deploy-version.rst
@@ -1,0 +1,68 @@
+========================
+Deploy Project Version
+========================
+
+**POST /api/v1/projects/:pluginId/versions/:version**
+
+Deploys a new version for the specified project. The body should be sent as ``multipart/form-data`` with the fields 
+shown below. Returns a JSON representation of the deployed version.
+
++-------------+------------------------------+----------------------------------------------------------------+
+| Name        | Data Type                    | Description                                                    |
++=============+==============================+================================================================+
+| apiKey      | String                       | An Ore deployment key obtained through the Ore settings.       |
++-------------+------------------------------+----------------------------------------------------------------+
+| channel     | String                       | The channel the version should be in.                          |
++-------------+------------------------------+----------------------------------------------------------------+
+| recommended | Boolean                      | If this version should be set as the recommended version.      |
+|             |                              | Defaults to true.                                              |
++-------------+------------------------------+----------------------------------------------------------------+
+| forumPost   | Boolean                      | Whether a forum post should be created or not. If no value is  |
+|             |                              | specified, it will default to the project's setting. Optional. |
++-------------+------------------------------+----------------------------------------------------------------+
+| changelog   | String                       | A changelog to include for this version. Optional.             |
++-------------+------------------------------+----------------------------------------------------------------+
+| pluginFile  | File                         | The plugin file to upload.                                     |
++-------------+------------------------------+----------------------------------------------------------------+
+| pluginSig   | File                         | A signature of the plugin file.                                |
++-------------+------------------------------+----------------------------------------------------------------+
+
+**Example curl usage:**
+
+.. code-block:: bash
+
+    curl -F "apiKey=string" -F "apiKey=string" -F "channel=string" -F "recommended=boolean" \
+    -F "forumPost=boolean" -F "changelog=string" -F pluginFile=@localPluginFile \
+    -F pluginSig=@localpluginSig https://URL.org/api/v1/projects/:pluginId/versions/:version
+
+**Example output:**
+
+.. code-block:: json
+
+    {
+        "id": 1226,
+        "createdAt": "2018-02-08 18:41:36.917",
+        "name": "1.1",
+        "dependencies": [{
+            "pluginId": "spongeapi",
+            "version": "7.0.0"
+        }],
+        "pluginId": "auction",
+        "channel": {
+            "name": "Release",
+            "color": "#009600"
+        },
+        "fileSize": 11441,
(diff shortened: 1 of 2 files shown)


source/ore/routes/deploy-version.rst
```diff
@@ -0,0 +1,57 @@
...
+        "tags": [{
+            "id": 13,
+            "name": "Sponge",
+            "data": "7.0.0",
+            "backgroundColor": "#F7Cf0D",
+            "foregroundColor": "#000000"
+        }],
+        "downloads": 26,
+        "author": "ewoutvs_"
+    }
```
ST-DDT: * [x] Newline at EOF.

source/ore/routes/deploy-version.rst
```diff
@@ -0,0 +1,57 @@
...
+========================
+
+**POST /api/v1/projects/:pluginId/versions/:version**
+
+Deploys a new version for the specified project. The body should be sent as ``multipart/form-data`` with the fields shown below. Returns a JSON representation of the deployed version.
+
++-------------+------------------------------+------------------------------------------------------------------------------------------------+
+| Name        | Data Type                    | Description                                                                                    |
++=============+==============================+================================================================================================+
+| apiKey      | String                       | An Ore deployment key gotten through the Ore settings.                                         |
```
Inscrutable: Can we use "obtained" instead of "gotten"?

ST-DDT (changes requested):
Could you add an example `curl` request which specifies the above parameter so its easier to decipher what you actually have to send. Because some APIs wrap their (meta)data in json.

source/ore/routes/deploy-version.rst
```diff
@@ -0,0 +1,57 @@
...
++-------------+------------------------------+------------------------------------------------------------------------------------------------+
+| channel     | String                       | The channel the version should be in.                                                          |
++-------------+------------------------------+------------------------------------------------------------------------------------------------+
+| recommended | Boolean                      | If this version should be set as the recommended version. Defaults to true.                    |
++-------------+------------------------------+------------------------------------------------------------------------------------------------+
+| forumPost   | Boolean                      | Overrides the project setting of if a forum post should be created for this version. Optional. |
```
Inscrutable: Does this really mean "If True, create a forum post for this version. This overrides the project setting." ?

source/ore/routes/deploy-version.rst
```diff
@@ -0,0 +1,57 @@
+========================
+Deploy Project Version
+========================
+
+**POST /api/v1/projects/:pluginId/versions/:version**
+
+Deploys a new version for the specified project. The body should be sent as ``multipart/form-data`` with the fields shown below. Returns a JSON representation of the deployed version.
+
++-------------+------------------------------+------------------------------------------------------------------------------------------------+
```
Inscrutable: Does the table really need to be this wide in raw format? Our preference is 120 character wide or lower. (Guideline 13 in https://docs.spongepowered.org/stable/en/contributing/spongedocs.html). It's easier to edit.

Inscrutable (commented):
Mostly makes sense. Table is big and ugly but not a critical issue.

source/ore/routes/deploy-version.rst
ST-DDT: Maybe we could phrase it like this. 
`Whether a forum post should be created or not. If no value is specified, it will default to the project's setting.`

Katrix:
Fixed problems and included suggestions. Not too familiar with Curl (generally just use Powershell) so it would be nice if someone else could fix that.

Inscrutable:
If someone can provide a curl example it would be handy, but otherwise I'm happy to merge this. Pending the addition of a final newline, of course.

phit:
this should work for a deployment, not tested
```
curl -F "apiKey=string" -F "apiKey=string" -F "channel=string" -F "recommended=boolean" \
-F "forumPost=boolean" -F "changelog=string" -F pluginFile=@localPluginFile \
-F pluginSig=@localpluginSig https://URL.org/api/v1/projects/:pluginId/versions/:version
```

Katrix:
Ok, added the curl usage. The newline should already be there.

source/ore/routes/deploy-version.rst
```diff
@@ -0,0 +1,68 @@
...
+| pluginSig   | File                         | A signature of the plugin file.                                |
++-------------+------------------------------+----------------------------------------------------------------+
+
+**Example curl usage:**
+
+.. code-block:: bash
+
+    curl -F "apiKey=string" -F "apiKey=string" -F "channel=string" -F "recommended=boolean" \
+    -F "forumPost=boolean" -F "changelog=string" -F pluginFile=@localPluginFile \
+    -F pluginSig=@localpluginSig https://URL.org/api/v1/projects/:pluginId/versions/:version
```
ST-DDT: Please replace `URL.org` with the real URL.

ST-DDT (changes requested):

ST-DDT:
I guess there a quite a few developers using Windows out there.

Maybe we should add an Powershell example as well? That way we could encourage new developers to try the API as well. What do you think?

phit:
shouldnt really be needed, https://blogs.technet.microsoft.com/virtualization/2017/12/19/tar-and-curl-come-to-windows/

ST-DDT:
@Katrix- AFAICT the url is the only remaining issue with this PR.
Could you please update it or is there anything else missing before merge?

Katrix:
Sorry, kind of forgot about that one. Fixed

ST-DDT (approved):

PR #671, opened 2018-03-10
New Terms Of Service

Inscrutable:
Created, formatted for readability and renumbered. More commits to come @Owexz

### DIFF
--- a/source/about/tos.rst
+++ b/source/about/tos.rst
@@ -4,10 +4,11 @@ Terms of Service
  
  .. code-block:: none
  
-    While our terms of service, privacy policy and community guidelines may be provided in your own language,
-    this is only as a courtesy to You. To the fullest extent as permitted by law, the governing language of
-    these documents is U.S. English. You may view the U.S. English version by clicking on the language
-    selector at the bottom right of the sidebar on this page and clicking on the U.S. flag.
+    While our terms of service, privacy policy and community guidelines may be provided in your
+    own language, this is only as a courtesy to You. To the fullest extent as permitted by law,
+    the governing language of these documents is U.S. English. You may view the U.S. English
+    version by clicking on the language selector at the bottom right of the sidebar on this
+    page and clicking on the U.S. flag.
 
 The following terms and conditions govern all use of the SpongePowered Services and all Content, Websites and products
 available at or through our services, including, but not limited to, SpongePowered Forum Software, (taken together,
--- a/source/about/tos.rst
+++ b/source/about/tos.rst
@@ -71,7 +71,7 @@ graphics, an audio file, or computer software. By making Content available, you 
   uses and effects of the materials, whether requested to do so by SpongePowered or otherwise.
 
 (3) User Content License
-~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~
 
 (3.1) All Platforms
 -------------------
--- a/source/about/tos.rst
+++ b/source/about/tos.rst
@@ -1,0 +1,285 @@
+Terms of Service
+================
+
+The following terms and conditions govern all use of the SpongePowered Services and all Content, Websites and products
+available at or through our services, including, but not limited to, SpongePowered Forum Software, (taken together,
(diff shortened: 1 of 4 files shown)


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/94b5704/.

- added: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/94b5704/about/privacy.html
- added: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/94b5704/about/tos.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/94b5704/about/index.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/94b5704/about/rules.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/671#issuecomment-371986419), not to the direct url to the preview._

source/about/tos.rst
```diff
@@ -0,0 +1,285 @@
+================
+Terms of Service
+================
+
+The following terms and conditions govern all use of the SpongePowered Services and all Content, Websites and products
+available at or through our services, including, but not limited to, SpongePowered Forum Software, (taken together,
+the Services). The Services are offered subject to your acceptance without modification of all of the terms and
+conditions contained herein and all other operating rules, policies (including, without limitation, SpongePowered’s
+Privacy Policy and Community Guidelines ) and procedures that may be published from time to time on this Site by
```
progwml6: both of these were linked, and both should be on the docs if not already on the docs

Inscrutable: The latter is already there, I can just add an internal link. 
Someone send me a copy of the ``Privacy Policy`` and we can shovel it in too.

ST-DDT (approved):
IANAL so I cannot verify the legal  part of this PR. But otherwise it looks good to me.

Maybe we should add a link / hint on where / how to  report content that does violate our TOS, any law or third party right. (Other than the DMCA)

dualspiral:
It just occurred to me that this will get translated. We **need** to add a provision that states that the English (US) version of the ToS is the prevailing version and that any translation into any other language is provided as a courtesy only.

If we do not do this, any translations that lag behind the ToS may be construed as the official language. We **want** the U.S. English version to be the governing document at all times, and not the ToS in the lagnuage that hasn't been updated for a few months. This statement **must** also be in English on _every translated page_ that contains our ToS, though a translated version can (and should) also be included. We should write something like:

"While our terms of service, privacy policy and community guidelines may be provided in your own language, this is only as a courtesy to You. To the fullest extent as permitted by law, the governing language of these documents is U.S. English. You may view the U.S. English version by clicking on the language selector at the bottom right of the sidebar on this page and clicking on the U.S. flag."

Inscrutable:
If I add it as a code block it will bypass translation. I could do that with the entire TOS ... 😈

source/about/tos.rst
progwml6: The link is in the original document

Inscrutable: "It was last updated May 31, 2013". If that's what we want, OK ...

Inscrutable:
_We've hammered out our policy
Agreed on the terms, more or less
Finally it's here for all to see
Here is your shiny new TOS_

Inscrutable:
It would be useful to cherry-pick this to the API 5 and 6 branches before we wrap them up totally.

Dannyps:
`If this server is in the USA, and you are under the age of 13, per the requirements of COPPA (Children's Online Privacy Protection Act), do not use this site.` https://github.com/SpongePowered/SpongeDocs/blame/stable/source/about/privacy.rst#L80

"If this server"? This seems to be a little out of context.

------------

`*This document is CC-BY-SA. It was last updated May 31, 2013.*`
https://github.com/SpongePowered/SpongeDocs/blame/stable/source/about/privacy.rst#L80

Perhaps there was a mistake here. There was no Sponge back in 2013.

Inscrutable:
Fixed, thanks for spotting those.

PR #677, opened 2018-04-05
Updated sample forge version to 1.12.2 to reflect API7

worm424:


ST-DDT (approved):
Are these all referenes to forge jars?
Or do we have the same issue with other versions in different places in the docs?

worm424:
I just did the forge jars on the launch script page,

Inscrutable:
There's another in ``server/getting-started`` which we can update, and fix the minimum java reference at the same time (from the OCD issue list).

phit:
doesn't Sponge need update 40? pretty sure update 20 will crash

Inscrutable:
This was added to the [OCD list](https://github.com/SpongePowered/SpongeDocs/issues/437#issuecomment-365637562) a while back. Gabi told me that the choice of version 40 was arbitrary. Feel free to check and confirm in case everything blows up.

stephan-gh:
Related: https://github.com/SpongePowered/SpongeCommon/commit/38a0642122d796de6413ff421848e84a094e401d

source/server/getting-started/implementations/spongeforge.rst
```diff
@@ -42,8 +42,8 @@ Example
 
 SpongeForge Jar files will always follow this naming scheme, to allow you to easily identify compatibility.
 
-For example the file name ``spongeforge-1.10.2-2254-5.2.0-BETA-2234.jar`` is compatible with Minecraft version
-``1.10.2``, was built with Forge ``12.18.3.2254`` (Build ``2254``), provides SpongeAPI ``5.2.0`` and was build number
+For example the file name ``spongeforge-1.12.2-2611-7.1.0-BETA-2990.jar`` is compatible with Minecraft version
+``1.12.2``, was built with Forge ``114.23.2.2611`` (Build ``2990``), provides SpongeAPI ``7.1`` and was build number
 ``2234`` of SpongeForge.
```
phit: `was built with Forge ``114.23.2.2611`` (Build ``2611``), provides SpongeAPI ``7.1.0`` and was build number
 ``2290`` of SpongeForge.`

phit (dismissed):

source/server/getting-started/implementations/spongeforge.rst
Inscrutable: That should be ``2990`` not ``2290`` ...

phit: numbers are hard!

PR #679, opened 2018-04-09
Update sponge-conf.rst

Leeo97one:
Fixed some defaults and added "max-chunk-unloads-per-tick".


Inscrutable:
Thanks for the update - but the bottom half says 
``This config was generated using SpongeForge build 2931 (with Forge 2611), SpongeAPI version 7.0.0`` - which version did you use? The raw config file below it needs updating too.

phit:
@Inscrutable the raw config is already up-to-date, I just missed those values at the top when doing the cleanup.

Leeo97one:
Ah, indeed, I did not pay attention to that. I just checked and everything is OK now.

PR #685, opened 2018-05-02
Update GAME_STOPPING and GAME_STOPPED

afriestad:
GAME_STOPPING and GAME_STOPPED apparently called the same events as SERVER_STOPPING and SERVER_STOPPED, respectively. I made an educated guess that the server states should call the server events and the game states should call the game events.


ST-DDT (approved):
Good catch.

Issue #689, opened 2018-05-08
Clarification of code sample licensing


parlough:
From the looks of it, even code snippets are licensed under `CC-BY-SA-4.0` which is not meant for code, and has some implications that basically makes a lot of plugins which may have taken code from the docs break the license. 

The code on the docs is meant for developers to use and not have to be worried about meeting some license requirements. I honestly think we should release all code samples/snippets on the docs under something like the [Unlicense](https://choosealicense.com/licenses/unlicense/) as developers shouldn't have to worry about including our copyright notice/MIT license for some code samples they took from our documentation. If this is not desired or too volatile, the MIT license will work as well.

Too avoid concerns with relicensing and getting agreements from everyone, I will be happy to rewrite every code sample/snippet from scratch(which has the added benefit of catching outdated code). 

This will also require us to include clarification on what is licensed under what and include the new license. I would handle that in my rewrite PR.

So I guess from here, I just need clarification on if the code's license is already MIT or something similar, and if not - what license we should go with for the code from here. Unlicense, MIT, custom license, your favorite permissive license goes here?

ST-DDT:
I'm willing to help you with this.

If we rewrite  code examples we should think of a way to detect scripts that no longer compile after a api update. So they won't get outdated again. Maybe we can do it as part of #690

Inscrutable:
It may seem odd, but SpongeDocs is effectively a *textbook* and the code examples are treated as such. They should be able to be re-used without license implications, under some interpretation of "Fair Use" (the right to use portions of copyrighted materials without permission for purposes of **education**, commentary, or parody). The usual acknowledgement for a written reference would apply in this case. i.e.:

"Code examples contained in SpongeDocs are provided as starting points for development, and are necessarily incomplete. We permit the use of these code examples in Sponge plugins under Fair Use, without further license restrictions, and we simply ask that SpongeDocs be acknowledged when this is the case." (stronger wording than "ask" may be needed).

That said, it would be preferable to have it all under **MIT** so there is no confusion between the examples and the licensing of SpongeAPI itself.

parlough:
Say we relicense the code examples under the MIT license, should we still include this fair use statement? I feel it still has value and our stance is still the same.

Inscrutable:
[...]

Issue #618, opened 2017-08-14
SpongeCommon/SpongeForge command changes


Grinch:
Just wanted to note a few changes to the `/sponge plugins` command.

As of now SpongeCommon uses pagination for the plugin command to display installed plugins. On SpongeCommon/SpongeVanilla it forces `Minecraft, MCP, SpongeAPI, and Sponge` at the beginning of the list in that order. The console will display a non-paginated list for this command.

SpongeForge filters out mods from the `/sponge plugins` command and added a new command `/sponge mods` which will display all mods. This forces `Minecraft, MCP, FML, MinecraftForge, SpongeAPI, and Sponge` at the beginning of the list in that order. The console also displays a non-paginated list for this command. This command uses the permission `sponge.command.mods`.

Both commands also make use of tooltips for each listed item that allow you to hover over the plugin/mod and see the ID and Version quickly. Clicking it still acts the same by showing a more detailed output of the plugin/mod's properties.

**tl;dr**: `/sponge plugins` now is a paginated list when sent to players. Filters out mods on SpongeForge. SpongeForge added `/sponge mods` which displays installed mods and uses the permission `sponge.command.mods`

Inscrutable:
Which versions of SpongeAPI does this affect (ie. is it being cherry-picked to older versions)?

Grinch:
The only version I targeted for this is 7. It could be pulled into older versions but that isn't something I personally plan to PR.

parlough:
I'll be working on an API 7 cleanup soon. Summer break here I come.

ryantheleach:
In addition:

https://docs.spongepowered.org/stable/en/server/spongineer/commands.html#sponge

Needs to be double checked against https://github.com/SpongePowered/SpongeCommon/blob/stable-7/src/main/java/org/spongepowered/common/command/SpongeCommandFactory.java#L175

The blockinfo and entityinfo are missing, maybe others?

ST-DDT:
@parlough Status?

ST-DDT:
@parlough Do you mind if I take over?

ST-DDT:
Note to self: The following commands are not or wrongly listed:

**Missing**

* /sponge blockinfo
* /sponge entityinfo
* /sponge which
* /callback

**Changed**

* /sponge help -> /help

Inscrutable:
The `callback` command is exclusively used by plugins, so that's probably all that needs to be said about it.

PR #691, opened 2018-05-11
[WIP] License clarification and rewrite code samples

### LINKED ISSUE #640: Outdated imports in docs

parlough:
I still have to rewrite most of the code samples but I have started setting up some of the changes and clarifying licenses. People can review as I go. 

In this PR I will also be updating other information as it stands in API 7.

Will address https://github.com/SpongePowered/SpongeDocs/issues/689 when finished.

This should also close https://github.com/SpongePowered/SpongeDocs/issues/640 and https://github.com/SpongePowered/SpongeDocs/issues/618

### DIFF
--- a/MIT-LICENSE
+++ b/MIT-LICENSE
@@ -1,0 +1,22 @@
+The MIT License (MIT)
+
+Copyright (c) SpongePowered <https://www.spongepowered.org>
+Copyright (c) contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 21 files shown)


parlough:
There's a long way to go, but @ST-DDT you should have push access if you ever have a chance to lend a hand. Don't really have a list to track what I've rewritten at the moment, but you can just look at files changed for now.

parlough:
Will need to modify the theme a little bit once this is done to add MIT to the footer - https://github.com/SpongePowered/sponge-docs-theme/blob/master/src/theme/templates/footer.html

Inscrutable:
There is also the possibility of using the Unlicense license for the code. http://unlicense.org/

ST-DDT:
Please resolve merge conflicts.

Inscrutable:
@parlough what has become of thee?

ST-DDT:
@parlough Can we merge most of this in, so that it does not gather any more dust? AFAIK everything can be merged except for the code samples license.

If you want I can give you a helping hand and fix the merge conflicts for you, then you just have to select the parts that can be merged earlier.

Inscrutable:
I'm afraid that @parlough has been unavailable/busy IRL for the last few months, so if you can clean this up @ST-DDT it would probably be the most expedient solution.

ST-DDT:
Superseded by #773

PR #728, opened 2018-08-03
Fix extra inline code block tick.

benjaminmordaunt:
This corrects an extra inline code tick, causing a backtick appearing within the code section.


ST-DDT (approved):

ST-DDT:
I also checked the rest of the docs, there are no further errors like this present.

PR #738, opened 2018-08-07
Add a note about thread safety. Fixes #242.

### LINKED ISSUE #242: Explain contexts.
Contexts are a core idea to implementing a permissions service, and is looking to also be used in a pending economy API.

As it is now available to use in other services, it might be worth explaining contexts, and how they can be used in plugins or custom services.

ST-DDT:
Fixes #242.

### DIFF
--- a/source/plugin/permissions.rst
+++ b/source/plugin/permissions.rst
@@ -290,8 +290,13 @@ context key with your plugin id) unless these contexts are meant to be shared.
 .. note::
     
     Please make sure that your ``ContextCalculator`` responds as **fast** as possible as it will get called frequently.
-    ``ContextCalculator`` implementations must be **thread safe**, because they might be called for async processing or
-    called in parallel.
+
+.. warning::
+
+    ``ContextCalculator`` implementations must be **thread safe**, because someone will be calling it from outside of
+    the main thread or even calling it in parallel. For to this reason, all non-name or non-uuid based 
+    ``ContextCalculator``\s (such as location based ones) have to utilize some kind of cache and must be to be updated
+    using event listeners or synchronized schedulers.
 
 Example
 ~~~~~~~
--- a/source/plugin/permissions.rst
+++ b/source/plugin/permissions.rst
@@ -289,7 +289,9 @@ context key with your plugin id) unless these contexts are meant to be shared.
 
 .. note::
     
-    Please make sure that your ``ContextCalculator`` responds as fast as possible as it will get called frequently.
+    Please make sure that your ``ContextCalculator`` responds as **fast** as possible as it will get called frequently.
+    ``ContextCalculator`` implementations must be **thread safe**, because they might be called for async processing or
+    called in parallel.
 
 Example
 ~~~~~~~
--- a/source/plugin/permissions.rst
+++ b/source/plugin/permissions.rst
@@ -293,10 +293,10 @@ context key with your plugin id) unless these contexts are meant to be shared.
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/956559d/.

- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/956559d/plugin/permissions.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/738#issuecomment-411131908), not to the direct url to the preview._

ryantheleach:
Needs to be stronger IMO. Most of the game thread in Minecraft is **not** thread safe, so there needs to be details regarding caching that data, and polling for it when it could change.

It may not be efficient, but it is correct, and SOMEONE will absolutely call permissions from off-thread and expect it to work.

ryantheleach (changes requested):
Needs to be stronger IMO. Most of the game thread in Minecraft is not thread safe, so there needs to be details regarding caching that data, and polling for it when it could change.

It may not be efficient, but it is correct, and SOMEONE will absolutely call permissions from off-thread and expect it to work, based on staff comments (if it's not explicitly written anywhere that permissions is threadsafe).

Happy to accept as-is though, but would like more.

ryantheleach:
A simple non-threadsafe context that could be used for example is "near spawn" for anti-spawn camping permissions.

The position of the player would need to be cached every tick, or maybe a move listener that determines where the player is. should be stored in a concurrent map of some sort, or lock on entries or something.

ST-DDT:
@ryantheleach Like this? Or should I add an explicit example of something bad + a potential fix?

ryantheleach:
LGTM;

"add an explicit example of something bad"

Merge it as-is ( I feel like I've already been a little fussy), if you feel like it, add an example of something **good** only.  I feel like Users will already know how to do something bad, and that offering them a bad code example will only make them think "I can't be bothered making it correctly" and copy the bad example because it looks simpler then the good example.

The note captures what needs to be said perfectly.

source/plugin/permissions.rst
```diff
@@ -289,7 +289,14 @@ context key with your plugin id) unless these contexts are meant to be shared.
 
 .. note::
     
-    Please make sure that your ``ContextCalculator`` responds as fast as possible as it will get called frequently.
+    Please make sure that your ``ContextCalculator`` responds as **fast** as possible as it will get called frequently.
+
+.. warning::
+
+    ``ContextCalculator`` implementations must be **thread safe**, because someone will be calling it from outside of
```
Inscrutable: It's a Plural. We also try to use passive third person, ie.
"because they may be called from outside of the main thread, or even called in parallel."

source/plugin/permissions.rst
```diff
@@ -289,7 +289,14 @@ context key with your plugin id) unless these contexts are meant to be shared.
...
+    the main thread or even calling it in parallel. For to this reason, all non-name or non-uuid based 
+    ``ContextCalculator``\s (such as location based ones) have to utilize some kind of cache and must be to be updated
```
Inscrutable: "... ones) have to utilize a cache, and must be updated ..."
(Do we really need "some kind of"?)

Inscrutable (changes requested):
I quibbled a couple of weird sentences. Good show.

Inscrutable (approved):

ryantheleach (approved):

PR #749, opened 2018-08-24
Document regex und selector support in CommandElements

### LINKED ISSUE #309: Coding commands for selectors. + regex!
```
9:59 AM <ryantheleach> If I want to add explicit selector support into my command, How can I accept one or many players as a command argument?
10:00 AM → Favorlock joined (~Favorlock@static-24-113-37-102.wavecable.com)
10:00 AM <+zml> ryantheleach: we're adding selector support to the player command element
10:00 AM <ryantheleach> so taking playerOrSource will in the future either return one, or a collection?
10:01 AM <ryantheleach> depending on how you get the args out of the context?
10:01 AM <+zml> in the future, will return a collection
10:01 AM <+zml> getOne will take the first element of a collection, returns absent if no or >1 elements
10:01 AM <+zml> use the onlyOne command element to enforce that count
10:01 AM <ryantheleach> Yup that's pretty much how I expected it to work, just was double checking
10:02 AM <ryantheleach> wait, hangon, if someone uses onlyOne, does that mean their command will get reprocessed for each player in the selector?
10:02 AM <+zml> no
10:03 AM <+zml> it means if the selector matches multiple players, it'll output an error message
10:04 AM <ryantheleach> Ok, so plugins must handle selectors explicitly in their code to have support. otherwise what will likely happen is they only teleport the first player in the list each time
10:04 AM <+zml> yeah
10:04 AM <ryantheleach> if it was a teleport command for instance
10:04 AM <+zml> also that command element supports matching players by regex
10:04 AM <+zml> as do a lot of sponge-provided arguments (like worlds, any cataloged type, etc)
New messages
10:05 AM <+zml> so you should be prepared to handle multiple arguments
10:05 AM <ryantheleach> by default? or plugins have to call a method on it to do that?
10:05 AM <+zml> by default
10:05 AM <ryantheleach> I'm going to make a docs issue, this needs to have more attention.
10:05 AM <+zml> It's in the javadocs I think :)
10:06 AM <+zml> but any command element implementation that extends PatternMatchingCommandElement can give you multiple results
```

So at the moment the default GenericArguments have support for regex selection of players, this should be explicitly mentioned in the docs, maybe in a warning or tip callout box. It's planned to shortly add support for selectors so selectors would also be handled in the same way.

ST-DDT:
Fixes #309

* Add explicit warning that there is regex and selector support there.
* Also use lambda for better readability
* Change title to `Plugin Commands` for disambiguation in the search.

### DIFF
--- a/source/plugin/commands/arguments.rst
+++ b/source/plugin/commands/arguments.rst
@@ -17,7 +17,7 @@ Argument Parsing
     java.lang.String
 
 The Command Builder API comes with a powerful argument parser. It converts the string input to java base types
-(integers, booleans, strings) or game objects (players, worlds, block types , ...). The parser supports optional
+(integers, booleans, strings) or game objects (players, worlds, block types, ...). The parser supports optional
 arguments and flags. It also handles TAB completion of arguments.
 
 The parsed arguments are stored in the :javadoc:`CommandContext` object. If the parser returns a single object, obtain
@@ -27,9 +27,9 @@ Many of the parsers may return more than one object (e.g. multiple players with 
 must use the :javadoc:`CommandContext#getAll(String)` method to get the ``Collection`` of possible matches.
 **Otherwise, the context object will throw an exception!**
 
-When creating a command, consider whether the argument could return multiple values, for example, whether a
-player argument could support multiple players when using a selector. If you support multiple values the users need
-to type only one command and can use an easier command syntax. Example: ``/tell @a Who took the cookies?``.
+When creating a command, consider whether the argument could return multiple values, for example, whether a player
+argument could support multiple players when using a selector. If you support multiple values the users need to type
+only one command and can use an easier command syntax. Example: ``/tell @a Who took the cookies?``
 
 .. tip::
 
@@ -41,7 +41,7 @@ To create a new :javadoc:`CommandElement` (argument), use the :javadoc:`GenericA
 elements require a short text key, which is displayed in error and help messages.
 
 Apply the ``CommandElement`` to the command builder with the :javadoc:`CommandSpec.Builder#arguments(CommandElement...)`
-method. It is possible to pass more than one ``CommandElement`` to the method, thus chaining multiple arguments (e.g
+method. It is possible to pass more than one ``CommandElement`` to the method, thus chaining multiple arguments (e.g.
 ``/msg <player> <msg>``). This has the same effect as wrapping the ``CommandElement`` objects in a
 :javadoc:`GenericArguments#seq(CommandElement...)` element.
 
--- a/source/plugin/commands/arguments.rst
+++ b/source/plugin/commands/arguments.rst
@@ -25,15 +25,17 @@ it with :javadoc:`CommandContext#getOne(String)`. Optional and weak arguments ma
 
 Many of the parsers may return more than one object (e.g. multiple players with a matching username). In that case, you
 must use the :javadoc:`CommandContext#getAll(String)` method to get the ``Collection`` of possible matches.
-**Otherwise, the context object will throw an exception!** Please make sure that you make a conscious decision whether
-you need to limit your ``CommandExecutor`` to a single result or can support multiple arguments. If you support multiple
-arguments the users have to type less and shorter commands. Example: ``/tell @a Who took the cookies?``.
+**Otherwise, the context object will throw an exception!**
+
+When creating a command, consider whether the argument could return multiple values, for example, whether a
+player argument could support multiple players when using a selector. If you support multiple values the users need
+to type only one command and can use an easier command syntax. Example: ``/tell @a Who took the cookies?``.
 
 .. tip::
 
-   You can use the :javadoc:`GenericArguments#onlyOne(CommandElement)` element to limit the amount of returned values to
-   a single one, so you can safely use ``args.<T>getOne(String)``. However the user will still get a message, if he
-   tries to select more than one value.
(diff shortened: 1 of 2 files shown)


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/950466d/.

- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/950466d/plugin/commands/arguments.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/950466d/plugin/commands/index.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/749#issuecomment-415844244), not to the direct url to the preview._

source/plugin/commands/arguments.rst
```diff
@@ -167,6 +166,12 @@ Overview of the ``GenericArguments`` command elements
 
     See the Javadocs for :javadoc:`GenericArguments` for more information.
 
+.. warning::
+
+    Some of these ``CommandElement``\s support multiple return values; some might event support regular expressions or
```
dualspiral: "event" -> "even"

Also, I would say that no command element should be expected to only return one element, and mention that if you want that guarantee of only one value being returned, it must be wrapped in `GenericArguments.onlyOne(...)`.

edit: you say that above, I see, but I'd reinforce it here too.

source/plugin/commands/arguments.rst
```diff
@@ -25,13 +25,15 @@ it with :javadoc:`CommandContext#getOne(String)`. Optional and weak arguments ma
...
+**Otherwise, the context object will throw an exception!** Please make sure that you make a conscious decision whether
+you need to limit your ``CommandExecutor`` to a single result or can support multiple arguments. If you support multiple
+arguments the users have to type less and shorter commands. Example: ``/tell @a Who took the cookies?``.
 
 .. tip::
 
-   You can use the
-   :javadoc:`GenericArguments#onlyOne(CommandElement)` element to limit the amount of returned values to a single one,
-   so you can safely use ``args.<T>getOne(String)``.
+   You can use the :javadoc:`GenericArguments#onlyOne(CommandElement)` element to limit the amount of returned values to
```
dualspiral: It doesn't _limit_ the number returned, that implies that if two values were to be returned, one would be instead. It ensures only one value can be returned, throwing an exception if more than one value would be returned.

I realise that the docs say that now, but we're better off correcting this while we're here.

source/plugin/commands/arguments.rst
```diff
@@ -25,13 +25,15 @@ it with :javadoc:`CommandContext#getOne(String)`. Optional and weak arguments ma
 
 Many of the parsers may return more than one object (e.g. multiple players with a matching username). In that case, you
 must use the :javadoc:`CommandContext#getAll(String)` method to get the ``Collection`` of possible matches.
-**Otherwise, the context object will throw an exception!**
+**Otherwise, the context object will throw an exception!** Please make sure that you make a conscious decision whether
```
Inscrutable: Not sure it needs to be this wordy. I think "You must decide whether you …" should do.
It also needs a comma and some word tweaking later "... single result, or allow it to support multiple arguments."

dualspiral: Make this point a separate paragraph. Something like:

> When creating a command, consider whether the argument could return more than multiple values, for example, whether a player argument could support multiple players when using a selector...

source/plugin/commands/arguments.rst
```diff
@@ -25,13 +25,15 @@ it with :javadoc:`CommandContext#getOne(String)`. Optional and weak arguments ma
...
+you need to limit your ``CommandExecutor`` to a single result or can support multiple arguments. If you support multiple
+arguments the users have to type less and shorter commands. Example: ``/tell @a Who took the cookies?``.
```
Inscrutable: "have to type less and shorter commands" could be "can use a shorter command syntax."

source/plugin/commands/arguments.rst
```diff
@@ -25,13 +25,15 @@ it with :javadoc:`CommandContext#getOne(String)`. Optional and weak arguments ma
...
 
 .. tip::
 
-   You can use the
-   :javadoc:`GenericArguments#onlyOne(CommandElement)` element to limit the amount of returned values to a single one,
-   so you can safely use ``args.<T>getOne(String)``.
+   You can use the :javadoc:`GenericArguments#onlyOne(CommandElement)` element to limit the amount of returned values to
+   a single one, so you can safely use ``args.<T>getOne(String)``. However the user will still get a message, if he
```
Inscrutable: he tries -> "they try" (genderless pronoun).

Inscrutable (commented):
Some streamlining.

source/plugin/commands/arguments.rst
ST-DDT: > return more than multiple values

Thats probably overkill... 
I decided to use `return multiple values` if I should use `return more than one value`  leave a comment below.

ST-DDT: This should be covered by dualspiral's suggestion

dualspiral: Ha, yeah, oops. Multiple is fine, IMO

source/plugin/commands/arguments.rst
ST-DDT: Fixed

PR #755, opened 2018-08-27
Start working on the metrics collection docs.

dualspiral:
After the merge of the Metrics Collection API for Ore, it's prudent that we document it!

I've added the basics so far, but I have more to do. Specifically:

* Add docs on what this section is in a bit more detail for a server admin, so they know that plugins downloaded from Ore have metrics collection off by default (maybe? Wanted? Where should it go?)
* Add docs on the API for plugin developers, including a link to the Ore guidelines (and maybe @mbax will want to chime in with thoughts on what we should add)

I'll do this over the next few days, uploading what I have now. Mostly want this visible so you know it's being worked on.

(Some line endings seem to have changed, probably because I'm on Windows at the moment. I'll have to look into that...)

PR #763, opened 2018-09-07
Preliminary guidelines update for 7.1.0

mbax:
Needs cleaning, URL to unfinished metrics docs, etc.

### DIFF
--- a/source/ore/guidelines.rst
+++ b/source/ore/guidelines.rst
@@ -2,10 +2,11 @@
 Ore Plugin Submission Guidelines
 ================================
 
-Welcome to the Ore submission guidelines. This document provides an outline of our expectations for both project and file submissions.
+Welcome to the Ore submission guidelines. This document provides an outline of our expectations for both project and
+file submissions.
 
-Remember that these are just guidelines and the Ore team, referred to as the “staff” throughout these guidelines, may choose to allow
-or disallow an action that is not explicitly listed here at our own discretion. 
+Remember that these are just guidelines and the Ore team, referred to as the “staff” throughout these guidelines, may
+choose to allow or disallow an action that is not explicitly listed here at our own discretion. 
 
 --------
 
@@ -19,8 +20,8 @@ Submitted projects should meet the following expectations:
 Name
 ~~~~
 
-Your project’s submitted name should not include a version, tagline, or other description. The name should be unique and original
-and must not have a name implying it is an official Sponge project.
+Your project’s submitted name should not include a version, tagline, or other description. The name should be unique
+and original and must not have a name implying it is an official Sponge project.
 
 Examples of names that are **not acceptable**:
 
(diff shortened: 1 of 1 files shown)


source/ore/guidelines.rst
```diff
@@ -160,6 +160,19 @@ system (such as a Sponge plugin that translates chat between languages), connect
 If your plugin sends information (e.g. a plugin list, player data, or map data) to external systems, the information collected
 must be listed on the main page (see above).
 
+    **Metrics (Data Collection)**
+
+    If any collected data (often referred to as "stats" or "metrics" data) about the server, such as version of server or plugins as
```
Inscrutable: Lines should have a maximum length of 120 characters (Docs Rule # 13). It makes proofreading easier.

source/ore/guidelines.rst
```diff
@@ -160,6 +160,19 @@ system (such as a Sponge plugin that translates chat between languages), connect
...
+    well as usage information, is to be sent to an external service the plugin must first query the Sponge API `MetricsConfigManager`.
```
Inscrutable: This first sentence is rather convoluted. May I suggest:
"Whenever data collected about the server (often referred to as "stats" or "metrics" data, such as server or plugin versions, as well as usage information) is to be sent to an external service, the plugin must first query the Sponge API `MetricsConfigManager`."

source/ore/guidelines.rst
```diff
@@ -160,6 +160,19 @@ system (such as a Sponge plugin that translates chat between languages), connect
...
+    Documentation on doing so can be found HERE [HALP, NEED LINK HERE]. This API must be checked before each time data is sent, not
```
Inscrutable: Dualspiral's Metrics PR is #755 (UPDATE: Now Merged). It creates the page `source/plugin/metrics.rst` which can be linked, if that is what you wish to refer to here.

Inscrutable: "This API must be checked before each time .." is a tad weird. Can we rearrange it or lose the "before"?

source/ore/guidelines.rst
```diff
@@ -160,6 +160,19 @@ system (such as a Sponge plugin that translates chat between languages), connect
...
+    only once. Plugins may not modify the values the API returns, but may encourage users to make the decision to enable the
+    collection and sending of this data for their plugin. The use of this API counts as the method by which to check if sending this
```
Inscrutable: Is there a better way to phrase this than "counts as the method"? It seems a little fuzzy to me.
You can probably use "metrics data" for "this data", as you have defined it above.

source/ore/guidelines.rst
```diff
@@ -160,6 +160,19 @@ system (such as a Sponge plugin that translates chat between languages), connect
...
+    data is enabled or disabled and no further check needs to be added.
+    
+    Note that this API was added in API 7.1.0 and so plugins build against
```
Inscrutable: Do you want to use a real Note?
I Think it's worth starting a new sentence after "API 7.1.0". 
"Plugins *built* against ... "
"... must instead check against ..." (can a plugin really have it's plugin do things?)

Inscrutable (commented):
The cat ate my homework.

ST-DDT:
@mbax Could you update/finish this one?

source/ore/guidelines.rst
mbax: Drop before is fine.

source/ore/guidelines.rst
mbax: "The use of this API satisfies the guidelines requirement of checking if the user has allowed sending metrics data."

Maybe?

source/ore/guidelines.rst
mbax: I'm worried about my own use of the English language here. Also, I believe a proper note could be useful.

source/ore/guidelines.rst
dualspiral: Is it even needed? Above, it states (emphasis mine):

> is to be sent to an external service the plugin **must** first query the Sponge API `MetricsConfigManager`

This would imply that is it mandatory to use this API in order to satisfy the guidelines, so this statement is unnecessary.

mbax: Fair point.

PR #771, opened 2018-10-20
Fix outdated causes in economy practices

ST-DDT:
Fix outdated causes as mentioned on Discord a few days ago.

### DIFF
--- a/source/plugin/economy/practices.rst
+++ b/source/plugin/economy/practices.rst
@@ -16,9 +16,9 @@ Withdrawing money
 Plugins should *not* check if an account has enough money before attempting to withdraw it. While this may
 sound counter-intuitive, it allows economy plugins to fully control how they handle negative balances.
 
-By checking yourself if the account has enough money, you prevent the economy plugin from (potentially)
-allowing a negative balance. For example, one economy plugin might want to allow negative balances to admins,
-or players with a certain permission. By performing the check yourself, you take this power away from the economy plugin.
+By checking yourself if the account has enough money, you prevent the economy plugin from (potentially) allowing a
+negative balance. For example, one economy plugin might want to allow negative balances to admins, or players with a
+certain permission. By performing the check yourself, you take this power away from the economy plugin.
 
 This code illustrates what **not** to do:
 
@@ -27,6 +27,8 @@ This code illustrates what **not** to do:
     import java.math.BigDecimal;
     
     import org.spongepowered.api.event.cause.Cause;
+    import org.spongepowered.api.event.cause.EventContext;
+    import org.spongepowered.api.event.cause.EventContextKeys;
     import org.spongepowered.api.service.economy.EconomyService;
     import org.spongepowered.api.service.economy.account.Account;
         
@@ -40,7 +42,7 @@ This code illustrates what **not** to do:
     } else {
         // The account has enough, let's withdraw some cash!
         account.withdraw(service.getDefaultCurrency(), requiredAmount,
-            Cause.source(this).build());
+            Cause.of(EventContext.of(ImmutableMap.of(EventContextKeys.PLUGIN, this)), this));
     }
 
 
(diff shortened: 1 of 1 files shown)


source/plugin/economy/practices.rst
```diff
@@ -16,9 +16,9 @@ Withdrawing money
 Plugins should *not* check if an account has enough money before attempting to withdraw it. While this may
 sound counter-intuitive, it allows economy plugins to fully control how they handle negative balances.
 
-By checking yourself if the account has enough money, you prevent the economy plugin from (potentially)
-allowing a negative balance. For example, one economy plugin might want to allow negative balances to admins,
-or players with a certain permission. By performing the check yourself, you take this power away from the economy plugin.
+By checking yourself if the account has enough money, you prevent the economy plugin from (potentially) allowing a
+negative balance. For example, one economy plugin might want to allow negative balances to admins, or players with a
+certain permission. By performing the check yourself, you take this power away from the economy plugin.
```
ST-DDT, line 21: Line length limit addressed.

source/plugin/economy/practices.rst
```diff
@@ -59,8 +61,8 @@ Here's how you **should** withdraw money:
     Account account = ...;
     BigDecimal requiredAmount = BigDecimal.valueOf(20);
 
-    TransactionResult result = account.withdraw(service.getDefaultCurrency(),
-        requiredAmount, Cause.source(this).build());
+    TransactionResult result = account.withdraw(service.getDefaultCurrency(), requiredAmount,
+                Cause.of(EventContext.of(ImmutableMap.of(EventContextKeys.PLUGIN, this)), this));
```
ST-DDT: It would be nice if I could get some feedback whether this is the expected way to specify the cause in this case. Or should more things added to the context/cause?

Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/52a725b/.

- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/52a725b/plugin/economy/practices.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/771#issuecomment-431612258), not to the direct url to the preview._

ST-DDT:
Some last change requests?

source/plugin/economy/practices.rst
ImMorpheus: You should be using the `PluginContainer`. 

As for the context, the alternative is 

`Cause.of(EventContext.builder().add(EventContextKeys.PLUGIN, container).build(), container);`

but that's just a matter of deciding which one is clearer. 

Sidenote: keep in mind that the builder is using an `HashMap` and the `EventContext` is using an `ImmutableMap`, so if you decide to use the builder you're iterating twice over the map.

See https://github.com/SpongePowered/SpongeAPI/blob/4f54ce0d8e487d139d3dcf9e39b29903c89b514b/src/main/java/org/spongepowered/api/event/cause/EventContext.java#L85 and https://github.com/SpongePowered/SpongeAPI/blob/4f54ce0d8e487d139d3dcf9e39b29903c89b514b/src/main/java/org/spongepowered/api/event/cause/EventContext.java#L224

ST-DDT: I like the builder suggestion.

Done

dualspiral (approved):
This looks fine as an alteration. 

A future PR could perhaps take note of the following: 

* The code block giving an example of what you shouldn't do has limited value as it's not translatable, it might be better to remove that and keep the statement (perhaps expanding upon it) that it is up to the economy plugin to determine whether a transaction succeeds, not the plugin.
* The causes are assuming that you're doing this async. It may be prudent to state that the economy API is intended to be thread safe. It may be an idea to link to the Causes page to explain how to create a cause off thread (or use the manager on thread).

12awsomeman34 (approved):

Issue #768, opened 2018-09-20
Changes coming to Versioning


Inscrutable:
We'll need to update the versioning page to deprecate the older Legacy system and add the new one when this PR goes live. Perhaps we can include some version of the tables there too.

https://github.com/SpongePowered/SpongeForge/issues/2367

ST-DDT:
This link contains some explanations that might help writing the docs: https://forums.spongepowered.org/t/sponge-status-update-10th-october-2018/25777

LordRalex:
The only page where I find the version references is on this page:
https://docs.spongepowered.org/stable/en/server/getting-started/implementations/spongeforge.html

Would it make more sense to have a dedicated version page that explained the format and have the install page reference that?

ST-DDT:
That is planned (as part of this issue). Maybe the info will be added to https://docs.spongepowered.org/stable/en/plugin/api-versions.html , but that is not sure. Or do you have a suggestion for an alternative location?

LordRalex:
That page would make the most sense, but the current dropdown location for it doesn't. I didn't find it because it's under a section I would not have thought to look. It also does really read more for development, so to me, it'd make sense to tailor it more as a general understanding of how the versions work and how to read it. Then the create and install pages can reference it directly. You don't see that page unless you click into the "Creating a Plugin" which for server administration, doesn't matter.

Grauldon:
The following pages also need updating:

- [https://docs.spongepowered.org/stable/en/contributing/versioning.html](https://docs.spongepowered.org/stable/en/contributing/versioning.html) 
- [https://docs.spongepowered.org/stable/en/plugin/internals/implementation.html](https://docs.spongepowered.org/stable/en/plugin/internals/implementation.html)

Here is a list of **all** the files that need to be changed:

source/server/getting-started/implementations/spongeforge.rst
source/plugin/api-versions.rst
source/plugin/internals/implementation.rst
source/contributing/versioning.rst

These files **can** be changed if the examples are changed. However, they specifically reference the `7.1.0` API, so they can remain as they are if preferred (maybe a note added to clarify the fact).

source/spongineer/logs.rst
source/spongineer/debugging.rst

> Would it make more sense to have a dedicated version page that explained the format and have the install page reference that?

[...]

PR #784, opened 2019-01-25
Begin debugging documentation

Grauldon:
This pull request is not complete and addresses issue #356 and PR #661.

Most changes are contained in the `source/contributing/implementation` directory. The one exception is a commit adding anchor points to the plugin debugging page in `source/plugin/debugging.rst`.

The branch began with a fork of PR #661. I decided on a new approach after reading comments in the PR. I changed `debugging.rst` to a directory so that content can expand over time without having one file too large or creating multiple debugging files in the implementation directory. Furthermore, I created stub files for future improvements based on comments in PR #661. 

I reviewed the SpongeDocs the other day for pertinent information and to evaluate the fit of the changes. It seems I am providing too many details for basic steps, and I need to ensure I am not providing information found elsewhere in the Docs. For example, some of the information in the `debugging/index.rst` file overlaps with the **Git Workflow for API and Implementations** page.

So, I decided it is time to reach out for assistance, guidance, and feedback. 

### Thoughts for moving forward

- Separate out command-line vs IDE versions of content (e.g. setup, debugging steps)
- Add subdirectories for certain content (e.g. setup via command line or IDE, mixin)
- Describe functionality in SpongeAPI vs SpongeCommon vs SpongeForge / SpongeVanilla
- Describe mixin, how it is integrated, debugging best practices and how-to's
- Complete stub pages (e.g. Debugging Mixin, Debugging Tools)
- Complete this "project" through multiple PR's by starting with basics and working through a prioritized list

### Questions

- Does what I have so far align well with the original ideas for this page?
- Are the pages too short (e.g. combine git, workspace, and ide)?
- What does this [comment](https://github.com/SpongePowered/SpongeDocs/pull/661#issuecomment-364664272) from PR 661 mean?

> Suggest the creation of a test plugin along with the contribution/PR.
- Can I get some input on the prioritized list (if not a provided list)?
- Are there other topics not listed here or in PR #661 that could be added?
- Can we promote the forum post entitled, [Common issues when debugging Sponge](https://forums.spongepowered.org/t/common-issues-when-debugging-sponge/26977)?

### Disclaimer

I am relatively new to Sponge and learning Java. I do have application development background, although on mainframes. So, please bear with me. I always heard that one way to learn processes and systems is to document them. Some of my documents may come across as too basic or like tutorials; this is why. It is also why I am reaching out for assistance. I want my documents to fit while providing pertinent information. All help and guidance is appreciated!

PR #785, opened 2019-02-04
Updates for new Versioning Policy

Grauldon:
This PR addresses SpongeDocs #768 and [SpongeForge Issue #2367](https://github.com/SpongePowered/SpongeForge/issues/2367).

It changes SpongeDocs by creating a new directory for the topic and adding pages explaining how Sponge manages versions. It also updates existing files to move version topics to one place within the SpongeDocs contents and adds references from existing locations to the new pages. The home page is also updated to reference the new pages.

The first commit updates SpongeDocs to provide the current Java version update shipped with the Mojang installer (1.8.0_74). This information is on the Minecraft Wiki for the Windows versions; I downloaded the Mac OS X  version and confirmed 1.8.0_74 is inside the .dmg file.

### DIFF
--- a/source/contributing/versioning.rst
+++ b/source/contributing/versioning.rst
@@ -14,7 +14,7 @@ The Bleeding Branch
 
 The core of our repositories is the ``bleeding`` branch. Almost all changes will be added to ``bleeding``, including
 new features, changes, and bugfixes. The version of ``bleeding`` will always be the next major release version 
-appended with ``-SNAPSHOT`` (e.g. ``8.0-SNAPSHOT``) to denote that it is not yet a final build and subject to change.
+appended with ``-SNAPSHOT`` (e.g. ``7.0.0-SNAPSHOT``) to denote that it is not yet a final build and subject to change.
 
 The primary reason for having the ``bleeding`` branch is to have a testing ground for changes. Even experienced
 members of the Sponge team can accidentally cause a build to fail or miss a bug. The ``bleeding`` branch will be
@@ -29,7 +29,7 @@ will be no breakages to API, only non-breaking additions. There is a branch name
 contains the latest API/implementation for that release including any minor or patch releases.
 
 When the time comes to release a major version, a new ``stable-x`` branch will be created from ``bleeding``, where
-``x`` is the new major version - for example, ``stable-8``. ``bleeding`` will be appropriately updated to be the next
+``x`` is the new major version - for example, ``stable-7``. ``bleeding`` will be appropriately updated to be the next
 major release as described above.
 
 Changes that have been in ``bleeding`` for a while, which have no known bugs, and that can be applied to a previous
--- a/source/contributing/versioning.rst
+++ b/source/contributing/versioning.rst
@@ -1,26 +1,9 @@
-==============================================
-Versioning System and Repository Branch Layout
-==============================================
+========================
+Repository Branch Layout
+========================
 
-With the release for beta we've moved SpongeAPI versioning to semantic versioning (see https://semver.org/).
-This change means that every time that we make a release we have to increment the version according to the rules
-of SemVer.
-
-SemVer
-======
-
-SemVer uses the scheme ``X.Y.Z``, where ``X`` is a *major* version, ``Y`` is a *minor* one and ``Z`` finally is a
-*patch* version.
-A release containing changes which are not backwards-compatible must be one *major* version ahead of the previous
-release. If there are only new features that are still backwards compatible, then the new release will be one *minor*
-version ahead of the previous release, and if the release strictly contains bugfixes then only the *patch* version will
-be incremented.
-
-This means that for example ``3.2.0`` is fully compatible to ``3.0.0`` while ``4.0.0`` isn't binary compatible to
-``3.0.0``. ``3.1.0`` and ``3.1.2`` are fully interchangeable besides the bugs that were fixed.
-
-The layout of our branches (described below) is designed to assist this process by allowing us to make minor releases
-without a breaking change forcing us to make it a major release. This branch layout applies to SpongeAPI,
+The layout of our branches is designed to assist :ref:`semantic versioning <sem-ver>` by allowing us to make minor 
+releases without a breaking change forcing us to make it a major release. This branch layout applies to SpongeAPI, 
 SpongeCommon, SpongeForge, and SpongeVanilla repositories but not to the SpongeDocs.
 
 SpongeAPI, SpongeCommon, SpongeForge and SpongeVanilla
@@ -30,8 +13,8 @@ The Bleeding Branch
 ~~~~~~~~~~~~~~~~~~~
 
 The core of our repositories is the ``bleeding`` branch. Almost all changes will be added to ``bleeding``, including
-new features, changes, and bugfixes. The version of ``bleeding`` will always be the next major release version
-appended with ``-SNAPSHOT`` (e.g. ``6.0.0-SNAPSHOT``) to denote that it is not yet a final build and subject to change.
+new features, changes, and bugfixes. The version of ``bleeding`` will always be the next major release version 
+appended with ``-SNAPSHOT`` (e.g. ``8.0-SNAPSHOT``) to denote that it is not yet a final build and subject to change.
 
 The primary reason for having the ``bleeding`` branch is to have a testing ground for changes. Even experienced
 members of the Sponge team can accidentally cause a build to fail or miss a bug. The ``bleeding`` branch will be
@@ -46,7 +29,7 @@ will be no breakages to API, only non-breaking additions. There is a branch name
 contains the latest API/implementation for that release including any minor or patch releases.
 
 When the time comes to release a major version, a new ``stable-x`` branch will be created from ``bleeding``, where
-``x`` is the new major version - for example, ``stable-5``. ``bleeding`` will be appropriately updated to be the next
+``x`` is the new major version - for example, ``stable-8``. ``bleeding`` will be appropriately updated to be the next
 major release as described above.
 
 Changes that have been in ``bleeding`` for a while, which have no known bugs, and that can be applied to a previous
--- a/source/plugin/api-versions.rst
+++ b/source/plugin/api-versions.rst
@@ -40,7 +40,7 @@ Getting the API Version from Implementations
 Getting it from the Jar
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-See :doc:`../../versions/versioning`.
+See :doc:`/versions/versioning`.
 
 .. _associated-minecraft-version:
 
--- a/source/plugin/api-versions.rst
+++ b/source/plugin/api-versions.rst
@@ -10,10 +10,13 @@ This page explains which API versions exist, and to which Minecraft version thei
 | *8.0.0*     | TBA          | TBA            | * *SpongeForge (1.13.x)*                  |
 |             |              |                | * *SpongeVanilla (1.13.x)*                |
 +-------------+--------------+----------------+-------------------------------------------+
+| 7.2.0       | TBA          | TBA            | * SpongeForge (1.12.2)                    |
+|             |              |                | * SpongeVanilla (1.12.2)                  |
++-------------+--------------+----------------+-------------------------------------------+
 | 7.1.0       | Sep 6, 2018  | TBA            | * SpongeForge (1.12.2)                    |
 |             |              |                | * SpongeVanilla (1.12.2)                  |
 +-------------+--------------+----------------+-------------------------------------------+
-| 7.0.0       | Jan 1, 2018  | TBA            | * SpongeForge (1.12.2)                    |
+| 7.0.0       | Jan 1, 2018  | Sep 5, 2018    | * SpongeForge (1.12.2)                    |
 |             |              |                | * SpongeVanilla (1.12.2)                  |
 +-------------+--------------+----------------+-------------------------------------------+
 | 6.0.0       | May 2, 2017  | Dec 31, 2017   | * SpongeForge (1.11.2)                    |
@@ -37,41 +40,7 @@ Getting the API Version from Implementations
 Getting it from the Jar
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-* For `SpongeForge <https://www.spongepowered.org/downloads/spongeforge/>`__: 
-
-  **Example:** ``spongeforge-1.12.2-2705-7.1.0-BETA-3361.jar``
-  
-  => API-Version ``7.1.0``
-  
-  See also :ref:`sponge-forge-file-name`.
-
-* For `SpongeVanilla <https://www.spongepowered.org/downloads/spongevanilla/>`__:
-
-  **Example:** ``spongevanilla-1.12.2-7.1.0-BETA-59.jar``
-  
-  => API-Version ``7.1.0``
-
-.. note::
-
-    The version string in SpongeForge & SpongeVanilla is currently flawed, it currently has a minor version number
-    referring to **unreleased** API versions, and is planned to be fixed going forward in some future update.
-
-    As we follow SemVer, changes in 7.1.0 should be backwards compatible with 7.0.0, as such, we often ship
-    SpongeForge & SpongeVanilla with **preview** implementations of the upcoming API release.
-
-    The API contained within the preview API builds can and will change before it's released, so for plugin
-    compatibility sake, it's best to use plugins for the latest stable released API, and not the API preview versions.
-
-    As such, if the API version number in the filename points to 7.2.0 but 7.2.0 is not yet announced to be released,
-    you should use plugins built against 7.1.0 unless specifically testing new functionality as asked by plugin devs, or
-    you are willing to put up with small breakages occasionally for the sake of new features.
-
-    This **WILL** be fixed in the near future. When this happens the Version number of SpongeForge & SpongeVanilla will
-    no longer reference unreleased SpongeAPI minor versions, but will still ship previews of the next minor API for
-    testing purposes.
-
-    Major API version bumps will be reserved for the experimental / bleeding builds, and will be very explicit as their
-    major version number will be different.
+See :doc:`../../versions/versioning`.
 
 .. _associated-minecraft-version:
 
--- a/source/plugin/internals/implementation.rst
+++ b/source/plugin/internals/implementation.rst
@@ -18,7 +18,7 @@ SpongeCommon
 
 - **Group ID**: ``org.spongepowered``
 - **Artifact ID**: ``spongecommon``
-- **Version**: Same as SpongeAPI
+- **Version**: Same as SpongeAPI + current recommended version (see :doc:`../../../versions/versioning` for details)
 - **Classifier**: ``dev``
 
 Example Using Gradle
@@ -27,7 +27,7 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongecommon:7.1.0:dev'
+        compile 'org.spongepowered:spongecommon:8.1.0:dev'
     }
 
 SpongeVanilla
@@ -46,7 +46,7 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongevanilla:1.12.2-7.1.0-BETA-116:dev'
+        compile 'org.spongepowered:spongevanilla:1.13.2-8.1.0-RC116:dev'
     }
 
 SpongeForge
@@ -65,5 +65,5 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongeforge:1.12.2-2705-7.1.0-BETA-3442:dev'
+        compile 'org.spongepowered:spongeforge:1.13.2-2705-8.1.0-RC3442:dev'
     }
--- a/source/plugin/internals/implementation.rst
+++ b/source/plugin/internals/implementation.rst
@@ -46,7 +46,7 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongevanilla:1.12.2-7.1.0-RC116:dev'
+        compile 'org.spongepowered:spongevanilla:1.12.2-7.1.1-RC123:dev'
     }
 
 SpongeForge
@@ -65,5 +65,5 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongeforge:1.12.2-2705-7.1.0-RC3442:dev'
+        compile 'org.spongepowered:spongeforge:1.12.2-2768-7.1.1-RC3484:dev'
     }
--- a/source/plugin/internals/implementation.rst
+++ b/source/plugin/internals/implementation.rst
@@ -27,7 +27,7 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongecommon:8.1.0:dev'
+        compile 'org.spongepowered:spongecommon:7.1.0:dev'
     }
 
 SpongeVanilla
--- a/source/plugin/internals/implementation.rst
+++ b/source/plugin/internals/implementation.rst
@@ -18,7 +18,7 @@ SpongeCommon
 
 - **Group ID**: ``org.spongepowered``
 - **Artifact ID**: ``spongecommon``
-- **Version**: Same as SpongeAPI + current recommended version (see :doc:`../../../versions/versioning` for details)
+- **Version**: Same as SpongeAPI + current recommended version (see :doc:`/versions/versioning` for details)
 - **Classifier**: ``dev``
 
 Example Using Gradle
--- a/source/plugin/internals/implementation.rst
+++ b/source/plugin/internals/implementation.rst
@@ -46,7 +46,7 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongevanilla:1.13.2-8.1.0-RC116:dev'
+        compile 'org.spongepowered:spongevanilla:1.12.2-7.1.0-RC116:dev'
     }
 
 SpongeForge
@@ -65,5 +65,5 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongeforge:1.13.2-2705-8.1.0-RC3442:dev'
+        compile 'org.spongepowered:spongeforge:1.12.2-2705-7.1.0-RC3442:dev'
     }
--- a/source/server/spongineer/debugging.rst
+++ b/source/server/spongineer/debugging.rst
@@ -23,7 +23,7 @@ Is the recommended Forge version installed?
 Usually SpongeForge will run on older or newer Forge builds than the recommended build.
 However, it is strongly advised to run the recommended build only.
 If you encounter a crash and your versions are mismatching, match them first and try again.
-If you're unsure which Forge build you need, take a look at :doc:`/server/getting-started/implementations/spongeforge`
+If you're unsure which Forge build you need, take a look at the **SpongeForge** section of :doc:`../../versions/filenames`.
 
 Are there any other coremods installed?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -95,7 +95,7 @@ Mismatched SpongeForge and Forge
 
 This is a common crash when you try to run SpongeForge on the wrong Forge build. Note that the target/Mixin can vary.
 Always match Forge against SpongeForge! If you're unsure which version of Forge is required and you already got your
-SpongeForge build, take a look at: :doc:`../getting-started/implementations/spongeforge/`
+SpongeForge build, take a look at the **SpongeForge** section of :doc:`../../versions/filenames`. 
 
 Other common errors
 ~~~~~~~~~~~~~~~~~~~
--- a/source/server/spongineer/debugging.rst
+++ b/source/server/spongineer/debugging.rst
@@ -23,7 +23,7 @@ Is the recommended Forge version installed?
 Usually SpongeForge will run on older or newer Forge builds than the recommended build.
 However, it is strongly advised to run the recommended build only.
 If you encounter a crash and your versions are mismatching, match them first and try again.
-If you're unsure which Forge build you need, take a look at the **SpongeForge** section of :doc:`../../versions/filenames`.
+If you're unsure which Forge build you need, take a look at the **SpongeForge** section of :doc:`/versions/filenames`.
 
 Are there any other coremods installed?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -95,7 +95,7 @@ Mismatched SpongeForge and Forge
 
 This is a common crash when you try to run SpongeForge on the wrong Forge build. Note that the target/Mixin can vary.
 Always match Forge against SpongeForge! If you're unsure which version of Forge is required and you already got your
-SpongeForge build, take a look at the **SpongeForge** section of :doc:`../../versions/filenames`. 
+SpongeForge build, take a look at the **SpongeForge** section of :doc:`/versions/filenames`. 
 
 Other common errors
 ~~~~~~~~~~~~~~~~~~~
--- a/source/server/spongineer/logs.rst
+++ b/source/server/spongineer/logs.rst
@@ -182,7 +182,7 @@ The example log indicates that we're running:
  [main/INFO] [FML]: Loading tweaker org.spongepowered.asm.launch.MixinTweaker from spongeforge-1.12.2-2705-7.1.0-BETA-3136.jar
 
 This indicates that SpongeForge 3136 was found and loaded by Forge. For further help regarding the SpongeForge
-naming scheme, have a look here: :doc:`../getting-started/implementations/spongeforge/`.
+naming scheme, have a look at :doc:`../../versions/filenames`.
 
 latest.log
 ~~~~~~~~~~
--- a/source/server/spongineer/logs.rst
+++ b/source/server/spongineer/logs.rst
@@ -182,7 +182,7 @@ The example log indicates that we're running:
  [main/INFO] [FML]: Loading tweaker org.spongepowered.asm.launch.MixinTweaker from spongeforge-1.12.2-2705-7.1.0-BETA-3136.jar
 
 This indicates that SpongeForge 3136 was found and loaded by Forge. For further help regarding the SpongeForge
-naming scheme, have a look at :doc:`../../versions/filenames`.
+naming scheme, have a look at :doc:`/versions/filenames`.
 
 latest.log
 ~~~~~~~~~~
--- a/source/versions/filenames.rst
+++ b/source/versions/filenames.rst
@@ -2,6 +2,11 @@
 Reading the Download Filename
 =============================
 
+.. note::
+
+    Our versioning policy was updated in October 2018. **Beginning with** SpongeAPI 7.1, the following information 
+    applies. See :doc:`legacy-versions` for information on older versions.
+
 When you download files, the name of the file will provide some important version information. The following sections 
 describe the information provided.
 
@@ -10,25 +15,14 @@ describe the information provided.
 SpongeForge
 ===========
 
-.. note::
-
-    Our versioning policy was updated in October 2018. **Beginning with** SpongeAPI 8.0, the following information 
-    applies. See :doc:`legacy-versions` for information on older versions.
-
-When you download SpongeForge, the name of the file includes a Forge build number which this version of SpongeForge is 
-built with and guaranteed to be compatible. We do not support newer versions of Forge; however, we try to get an update 
-out when a Forge release breaks SpongeForge. You can try the Forge version needed as it will work fine most of the 
-time. If it doesn't, we hope to release a SpongeForge compatible with that Forge version soon. 
-
-The format of the filename is:
+The format of the SpongeForge download filename is:
 
 ``spongeforge-<MCVersion>-<ForgeBuild>-<APIMajor>.<LatestAPIMinorRelease>.<RecommendedVersion(-RC<BuildNumber>)>.jar``
 
 +----------------------+-----------------------------------------------------------------------------------------------+
 | ``MCVersion``        | The Minecraft version. Only clients compatible with this version can connect.                 |
 +----------------------+-----------------------------------------------------------------------------------------------+
-| ``ForgeBuild``       | Preferably your server should run this exact version of Forge (which can be identified in the |
-|                      | last part of Forge's version string).                                                         |
+| ``ForgeBuild``       | The Forge build that SpongeForge is built against and is guaranteed to be compatible          |
 +----------------------+-----------------------------------------------------------------------------------------------+
 | ``APIMajor``         | The major version of SpongeAPI implemented by this file (the ``X`` in                         |
 |                      | :ref:`semantic versioning<sem-ver>`).                                                         |
@@ -43,8 +37,20 @@ The format of the filename is:
 | ``<BuildNumber>``    | present, the ``RecommendedVersion`` has **not** been released yet.                            |
 +----------------------+-----------------------------------------------------------------------------------------------+
 
+The Forge Build in the filename specifies the version of Forge this version of SpongeForge used during development and 
+testing. The two versions are guaranteed to work without any issues. We **tend** to use the latest *Recommended Build* 
+of Forge for this purpose.
+
+.. note::
+
+    Normal Forge mods can usually run on any build of Forge for a given Minecraft version (e.g. 1.12.2) without any 
+    problem. However, SpongeForge needs to access, among other things, internal parts of Forge, which most mods
+    shouldn’t be touching, let alone modifying as Sponge does. Since Forge is free to change their internal code
+    whenever they want to, its normal guarantee of backwards-compatibility doesn’t apply to SpongeForge. Feel free to
+    use more recent versions of Forge than the one used for SpongeForge, but we can't always guarantee compatibility.
+
 Examples
-~~~~~~~~
+--------
 
 SpongeForge Jar files will always follow this naming scheme to allow you to easily identify compatibility.
 
@@ -56,22 +62,6 @@ Another example is the file name ``spongeforge-1.12.2-2705-7.1.5-RC3449.jar``. T
 version ``1.12.2``, was built with Forge ``14.23.4.2705`` (Build ``2705``), guarantees compatibility with SpongeAPI 
 ``7.1.0``, is not a recommended version, and is build number ``3449`` in development, which will be the ``5`` th 
 release of SpongeForge once this version is released. 
-
-.. note::
-
-    Normal Forge mods can usually run on any build of Forge for a given Minecraft version (e.g. 1.8.0) without any
-    problem. However, SpongeForge needs to access, among other things, internal parts of Forge, which most mods
-    shouldn’t be touching, let alone modifying as Sponge does. Since Forge is free to change their internal code
-    whenever they want to, its normal guarantee of backwards-compatibility doesn’t apply to SpongeForge. Feel free to
-    use more recent versions of Forge, than the one used for SpongeForge, but we can't always guarantee compatibility.
-
-.. warning::
-
-    When investigating crash issues, you can freely try newer versions of Forge than listed on the SpongeForge Jar.
-    However, it is recommended to also check with the matching version, to make sure your issue is not related to a
-    version mismatch.
-    Even though there will be no guarantee of compatibility, please report any breakage to the issue tracker, so that
-    SpongeForge can be updated.
 
 SpongeVanilla
 =============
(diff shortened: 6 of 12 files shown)


source/contributing/versioning.rst
```diff
@@ -30,8 +13,8 @@ The Bleeding Branch
 ~~~~~~~~~~~~~~~~~~~
 
 The core of our repositories is the ``bleeding`` branch. Almost all changes will be added to ``bleeding``, including
-new features, changes, and bugfixes. The version of ``bleeding`` will always be the next major release version
-appended with ``-SNAPSHOT`` (e.g. ``6.0.0-SNAPSHOT``) to denote that it is not yet a final build and subject to change.
+new features, changes, and bugfixes. The version of ``bleeding`` will always be the next major release version 
+appended with ``-SNAPSHOT`` (e.g. ``8.0-SNAPSHOT``) to denote that it is not yet a final build and subject to change.
```
phit: should still be 8.0.0-SNAPSHOT

source/plugin/internals/implementation.rst
```diff
@@ -27,7 +27,7 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongecommon:7.1.0:dev'
+        compile 'org.spongepowered:spongecommon:8.1.0:dev'
```
phit: why change these to versions that dont exist yet?

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
+=============================
+Reading the Download Filename
+=============================
+
+When you download files, the name of the file will provide some important version information. The following sections 
+describe the information provided.
+
+SpongeAPI
```
phit: move spongeforge and vanilla above api

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
...
+    for an explanation of the ``-shaded`` label.
+
+.. _sponge-forge-file-name:
+
+SpongeForge
+===========
+
+.. note::
+
+    Sponge changed our versioning policy in October 2018. **Beginning with** SpongeAPI 8.0, the following information
```
phit: > Our versioning policy was updated in October 2018

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
...
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``LatestAPI``        | The minor version of SpongeAPI implemented by this file (the ``Y`` in                         |
+| ``MinorRelease``     | :ref:`semantic versioning<sem-ver>`).                                                         |
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``Recommended``      | The released version of the implementation when **not** followed by ``-RC<BuildNumber>.``     |
+| ``Version``          | When this number is **not** zero, the SpongeAPI version in development is used (i.e. API      |
+|                      | Minor + 1).                                                                                   |
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``-RC``              | The build number in development for the next implementation release. When a build number      |
```
phit: implementation > recommended

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
...
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``APIMajor``         | The major version of SpongeAPI implemented by this file (the ``X`` in                         |
+|                      | :ref:`semantic versioning<sem-ver>`).                                                         |
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``LatestAPI``        | The minor version of SpongeAPI implemented by this file (the ``Y`` in                         |
+| ``MinorRelease``     | :ref:`semantic versioning<sem-ver>`).                                                         |
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``Recommended``      | The released version of the implementation when **not** followed by ``-RC<BuildNumber>.``     |
+| ``Version``          | When this number is **not** zero, the SpongeAPI version in development is used (i.e. API      |
```
phit: I mean I guess, I would just leave this sentence out.

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
...
+| ``<BuildNumber>``    | is present, the ``RecommendedVersion`` has **not** been released yet.                         |
++----------------------+-----------------------------------------------------------------------------------------------+
+
+Examples
+~~~~~~~~
+
+SpongeForge Jar files will always follow this naming scheme to allow you to easily identify compatibility.
+
+For example, the file name ``spongeforge-1.12.2-2705-7.1.4.jar`` is compatible with Minecraft version ``1.12.2``, was 
+built with Forge ``14.23.4.2705`` (Build ``2705``), uses SpongeAPI ``7.2-SNAPSHOT``, is a recommended version, and is
```
phit: > uses SpongeAPI ``7.2-SNAPSHOT`` 

should be

> guarantees compatibility with SpongeAPI ``7.1.0``

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
...
+
+When you download files, the name of the file will provide some important version information. The following sections 
+describe the information provided.
+
+SpongeAPI
+=========
+
+The format of the filename is:
+
+``spongeapi-<APIMajor>.<LatestAPIMinorRelease>-shaded.jar``
```
phit: ``spongeapi-<APIMajor>.<LatestAPIMinorRelease>.0-shaded.jar``

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
...
+
+**OR**
+
+``spongeapi-<APIMajor>.<LatestAPIMinorRelease>.<Date>.<Time>-<BuildNumber>-shaded.jar``
```
phit: ``spongeapi-<APIMajor>.<NextAPIMinorRelease>.0-<Date>.<Time>-<BuildNumber>-shaded.jar``

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
...
+~~~~~~~~
+
+SpongeForge Jar files will always follow this naming scheme to allow you to easily identify compatibility.
+
+For example, the file name ``spongeforge-1.12.2-2705-7.1.4.jar`` is compatible with Minecraft version ``1.12.2``, was 
+built with Forge ``14.23.4.2705`` (Build ``2705``), uses SpongeAPI ``7.2-SNAPSHOT``, is a recommended version, and is 
+the ``4`` th release of SpongeForge with this API.
+
+Another example is the file name ``spongeforge-1.12.2-2705-7.1.5-RC3449.jar``. This file is compatible with Minecraft 
+version ``1.12.2``, was built with Forge ``14.23.4.2705`` (Build ``2705``), uses SpongeAPI ``7.2-SNAPSHOT``, is not a
```
phit: same as above for api

source/versions/legacy-versions.rst
```diff
@@ -0,0 +1,39 @@
+=============================
+Older Versions of SpongeForge
+=============================
+
+.. note::
+
+    Sponge changed our versioning policy in October 2018. **Prior to** SpongeAPI 8.0, the following information
```
phit: > Our versioning policy was updated in October 2018. **Prior to** SpongeAPI 8.0.0, ....

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
...
+
+SpongeForge
+===========
+
+.. note::
+
+    Sponge changed our versioning policy in October 2018. **Beginning with** SpongeAPI 8.0, the following information 
+    applies. See :doc:`legacy-versions` for information on older versions.
+
+When you download SpongeForge, the name of the file includes a Forge build number which this version of SpongeForge is
```
phit: I know this is old, but this isn't really true, would be nice if you can update this. The version in the filename is the version SpongeForge was built with, that guarantees compatibility, however most of the time newer Forge version also work fine but are not officially supported. So basically just word it differently, same applies to the same section on the old versioning page.

source/versions/versioning.rst
```diff
@@ -0,0 +1,118 @@
...
+| ``8.5.0`` | ``8.5`` | ``0``          |
+|           |         |                |
++-----------+---------+----------------+
+
+There are two other components of the version string: *SNAPSHOT* and *RC*. Both of these labels indicate a version 
+still in development. These files are used for testing new features and bug fixes, so they are generally used by 
+developers. *SNAPSHOT* is associated with the API, and *RC*, or Release Candidate, is associated with the 
+implementation (SpongeVanilla and SpongeForge).
+
+*SNAPSHOT* and *RC* are **always** associated with a number (e.g., ``1-SNAPSHOT``, ``-RC120``). With *SNAPSHOT*, the
```
phit: ``1-SNAPSHOT`` ?
what does that mean, snapshot is a suffix for the whole version before that, you cant just take the patch version by itself.
7.1.0-snapshot unlike RC120 is also not the name for a specific version, its refers to every build between releases.

source/versions/versioning.rst
```diff
@@ -0,0 +1,118 @@
...
+    Keep in mind the diffences between a major release, a minor release, and a patch release while reading the 
+    following.
+
+API
+---
+
+An important part to grasp in understanding Sponge versioning is how the version string conveys the API information. 
+The API version is primarily conveyed by the *major* . *minor* labels. The *major* label is always a number. The 
+*minor* label can be just a number, indicating a release, or it can be a number with ``-SNAPSHOT`` appended to it, 
+indicating a **minor release in development**. Developers are the most likely to use snapshots, so most Sponge users
```
phit: > Developers are the most likely to use snapshots

hopefully not, devs should use the latest stable release, snapshot features may break at any time if they arent part of the last release

source/versions/versioning.rst
```diff
@@ -0,0 +1,118 @@
...
+introduced by ``3.1.0``.
+
+Version String
+==============
+
+Applying the versioning policy to Sponge begins with SpongeAPI. It is the core on which the version string is based. 
+SpongeAPI is the actual API that allows plugins to interact with the game. However, the API is enabled by the Sponge 
+implementation (SpongeForge or SpongeVanilla). Together, these products form the foundation of a Sponge server.
+
+As such, the version string reflects the state of these two products. The major and minor release components of a
```
phit: I don't agree with this, the implementation does derive part of its version from the API for clarity sake, however the versions are disconnected. `8.0.2` will never be a thing in API and the part derived from the API is `8.0.0` not `8.0`.

So a lot on this page does not make much sense to me personally, how it is explained and how you split the version string, but maybe I'm seeing it wrong. We will see what others say, just leave it for now.

source/versions/filenames.rst
```diff
@@ -0,0 +1,121 @@
...
+.. note::
+
+    Sponge changed our versioning policy in October 2018. **Beginning with** SpongeAPI 8.0, the following information 
+    applies. See :doc:`legacy-versions` for information on older versions.
+
+When you download SpongeForge, the name of the file includes a Forge build number which this version of SpongeForge is 
+compatible with. Other builds, even ones differing by only a few build numbers, are not officially supported.
+
+However, SpongeForge usually updates to a new Forge build fairly soon after it's released, so you needn't worry about 
+always having to run an outdated Forge version in order to use SpongeForge.
```
Inscrutable: I'm not sure this is really true, or useful (reflecting phit's comments above). When a Forge release breaks SpongeForge, we try to get an update out, but otherwise it's a bit ad-hoc. Normally updating Forge is a safe operation, but there is no particular rule I can think of that covers it nicely. Other than "try it and see what happens".

source/versions/versioning.rst
```diff
@@ -0,0 +1,118 @@
...
+Applying the versioning policy to Sponge begins with SpongeAPI. It is the core on which the version string is based. 
+SpongeAPI is the actual API that allows plugins to interact with the game. However, the API is enabled by the Sponge 
+implementation (SpongeForge or SpongeVanilla). Together, these products form the foundation of a Sponge server.
+
+As such, the version string reflects the state of these two products. The major and minor release components of a 
+version is based on the API while the patch component of the version is based on the implementation. Let's consider 
+this version string as an example: ``8.2.0``. This string conveys the API is Version 8.2 and the implementation is the 
+first release using this API version.
+
+The following table provides several examples of a version string and its components:
```
Inscrutable: This table does not help much - it may actually add to confusion. I'm not sure dissecting the semver this way is really assisting. Maybe directing via to a use-case might improve it.

source/versions/versioning.rst
Inscrutable: It's often the way with other APIs, Devs will target new features they wanted, even if unstable, so SNAPSHOTS will still get used now and then. This is very much at their own risk, so development against a Recommended Build is preferred and encouraged.

Inscrutable (commented):
Excellent. I'll see what others have to add to this. We want it to be concise.

source/versions/versioning.rst
ImMorpheus: What about linking to the website like github does ? 
https://semver.org/

ImMorpheus:
@Grauldon Thanks for your contribution.

Just a note: the stable branch is used for the [7.X.Y version of the docs](https://docs.spongepowered.org/stable/en/).
For consistency and to avoid confusion once there's a dedicated docs for API 8, it's better to keep any version reference to API 7. Especially in `Example Using Gradle`.

Grauldon:
I'll plan to return to this PR this week. My original plan was to balance time between this one and #784 better. Using git stash to rebase with stable has helped me to realize how to accomplish this. Thanks for all of the feedback!

source/contributing/versioning.rst
Grauldon: Updated to 7.0.0-SNAPSHOT

source/plugin/internals/implementation.rst
Grauldon: Reverted version back to 7.1.0:dev.

source/versions/filenames.rst
Grauldon: Done

source/versions/filenames.rst
Grauldon: Changed

source/versions/filenames.rst
Grauldon: Done

source/versions/filenames.rst
Grauldon: Changed here and in the following example.

source/versions/filenames.rst
Grauldon: Done

source/versions/filenames.rst
Grauldon: Done

source/versions/filenames.rst
Grauldon: Done

source/versions/legacy-versions.rst
Grauldon: Done

source/versions/filenames.rst
Grauldon: Removed

source/versions/filenames.rst
Grauldon: Done

source/versions/filenames.rst
Grauldon: I used phit's comment to rewrite those paragraphs. We may need to make more changes after you review the new update.

source/versions/versioning.rst
```diff
@@ -0,0 +1,118 @@
+==========
+Versioning
```
dualspiral, line 2: First, please let me thank you for all the work you've done so far. I do not want to detract from it and this thought process would not exist if you hadn't stepped up, so thank you!

However, this particular page doesn't feel like it's particularly useful in this form. We talk about sem-ver, what it means, then the version string, but because there is so much on here, when you look at the API version string, for example, people may have lost the link to sem ver. It is better to explain what each bit means inline for us.

I think what I say needs discussion (@SpongePowered/docs) and this is just my own opinion, but I would:

* Remove the semver section, with the exception of a tip at the top saying "Sponge follows a format of the SemVer specification in its projects. You can read about general SemVer usage at http://semver.org". There really is no point in us rehashing it.
* Say that SpongeAPI and SpongeForge/SpongeVanilla follow two different versioning policies
* Highlight the important parts of the version string as the semver part and, for impl, the minecraft version
* Explain what that means for the developer (for the API version string) and the server owner (for the impl version string) - most of this page doesn't really decribe that and that's what the versioning is really all about for us.

For the API, it needs to boil down to this:

* A change in the major number indicates that some APIs that were guaranteed in the previous version have been broken. 5->6 broken world pregen APIs, for example. 6->7 was a more significant break. A bump in the major number is always accompanied by resettign the minor number to zero.
  * Developers know that their plugins _might_ break when this changes, i.e. API 6.x plugins might not run on API 7.x.
* A change in the minor number (but not the major number) means that API calls have been added, but plugins for previous minor revisions of the same major revision will still work - _but this isn't necessarily the case the other way around_.
  * Developers know their plugins will still function on 7.x+n (target 7.x, n≥0, so 7.0 plugins work on 7.1), but not necessarily the other way around.
* `-SNAPSHOT` means that the versioned API is still in progress
  * If the minor version is 0 (e.g. 8.0-SNAPSHOT), **anything** may break before it is released and plugins cannot expect stability.
  * If the minor version is greater than zero (e.g. 7.2-SNAPSHOT), anything up to the previous minor revision is still guaranteed (in this case, 7.1), but anything added since (in this case, for what will be 7.2) **may still break**, so cannot expect stability on unreleased features.

For the impl:

* A version string with an RC identifier means "recommended build", builds we consider as a reasonable quality that we would want to run ourselves.
* We tend to bump the build to indicate bug fixes and performance improvements, or any significant changes in the config or other systems that aren't related to the API (in this sense, the "patch" part of the version string isn't really a "patch" part and so it kind of moves away from the sem-ver ideology).
* For the 7.x.y part, 7.x is the version of the API that we guarantee for plugins, y is the recommended build version for this API series.
* Plugins written for 7.0 to 7.x will work. 7.x+n (n≥1) (including any SNAPSHOT features) _may not_ work.
* Any server software that has the version `n.0.0-SNAPSHOT` where the minor and recommended build numbers are 0 (or would the recommended build release be 1, I'd have to check) and the SNAPSHOT monkier is applied, versions are bleeding builds for testing the next version of the API only and may break at any time as we refine everything about it.

source/versions/filenames.rst
Grauldon: I realized you were mentioning the paragraph after phit's comment. So, I re-worked the changes to handle both comments. We still may need to make further changes, but I think you will like the new changes.

source/versions/versioning.rst
Grauldon: Thank you for the comment and feedback! This in-depth explanation provides an outline I can follow to get it right this time. I figured a re-write was needed; I just wasn't sure yet how to proceed. Your feedback will save me much time. Thanks again!

Grauldon:
I have updated the PR to address most of the reviews provided. Thanks to all for the feedback and pointing out mistakes and better ways of writing the documentation.

I did not work with versioning.rst. I had planned on spending much more time with it, and I have run out of time today. So, I am pushing the other changes I have made and will continue with the versioning.rst file Saturday. 

@ImMorpheus Thank you for the comment and feedback. The link is near the top. But like @dualspiral pointed out, the deluge of information drowns it out. Also, I made the corrections to `Example Using Gradle`. 

I will be away from my computer tomorrow, so I won't be able to make any updates until Saturday. I should be able to look over the feedback about versioning.rst some tomorrow though. I will push those changes Saturday evening if possible.

source/versions/versioning.rst
dualspiral: To be honest, what you had wasn't bad, it just seemed to kind of meander a bit much. You were doing a good job of being descriptive, just in the wrong areas! Regardless, I know that large comments that suggest a rewrite can be hard to take, so thank you for taking it in the good faith you have! :)

source/versions/filenames.rst
```diff
@@ -0,0 +1,117 @@
...
+===========
+
+.. note::
+
+    Our versioning policy was updated in October 2018. **Beginning with** SpongeAPI 8.0, the following information 
+    applies. See :doc:`legacy-versions` for information on older versions.
+
+When you download SpongeForge, the name of the file includes a Forge build number which this version of SpongeForge is 
+built with and guaranteed to be compatible. We do not support newer versions of Forge; however, we try to get an update 
+out when a Forge release breaks SpongeForge. You can try the Forge version needed as it will work fine most of the
```
ImMorpheus: This isn't true either.
SpongeForge is built against the latest **recommended** version of Forge, we don't bump the Forge version unless there's a new recommended build. 

See https://github.com/SpongePowered/SpongeForge/issues/2416 https://github.com/SpongePowered/SpongeForge/issues/2370

ImMorpheus (changes requested):

source/versions/filenames.rst
```diff
@@ -0,0 +1,117 @@
...
+    shouldn’t be touching, let alone modifying as Sponge does. Since Forge is free to change their internal code
+    whenever they want to, its normal guarantee of backwards-compatibility doesn’t apply to SpongeForge. Feel free to
+    use more recent versions of Forge, than the one used for SpongeForge, but we can't always guarantee compatibility.
+
+.. warning::
+
+    When investigating crash issues, you can freely try newer versions of Forge than listed on the SpongeForge Jar.
+    However, it is recommended to also check with the matching version, to make sure your issue is not related to a
+    version mismatch.
+    Even though there will be no guarantee of compatibility, please report any breakage to the issue tracker, so that
```
ImMorpheus, line 57: See above.

This is misleading, if you're not using the recommended forge version then there's nothing to fix as we won't break compatibility with the recommended build.

ImMorpheus (changes requested):

source/versions/filenames.rst
dualspiral: That said, we've also discussed updating to non-recommended builds, while it doesn't happen now, we should not be boxing ourselves into such a policy in future (and we should avoid making guarantees to something like this), especially as we will probably not think to update a sentence like this after the fact.

I would just delete the sentences "We do not support... breaks SpongeForge" and "If it doesn't, we... version soon" (but consider my other notes).

source/versions/filenames.rst
```diff
@@ -0,0 +1,117 @@
...
+describe the information provided.
+
+.. _sponge-forge-file-name:
+
+SpongeForge
+===========
+
+.. note::
+
+    Our versioning policy was updated in October 2018. **Beginning with** SpongeAPI 8.0, the following information
```
dualspiral: We started this with Sponge API 7.1+ builds

source/versions/filenames.rst
```diff
@@ -0,0 +1,117 @@
...
+time. If it doesn't, we hope to release a SpongeForge compatible with that Forge version soon. 
+
+The format of the filename is:
+
+``spongeforge-<MCVersion>-<ForgeBuild>-<APIMajor>.<LatestAPIMinorRelease>.<RecommendedVersion(-RC<BuildNumber>)>.jar``
+
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``MCVersion``        | The Minecraft version. Only clients compatible with this version can connect.                 |
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``ForgeBuild``       | Preferably your server should run this exact version of Forge (which can be identified in the |
```
dualspiral: "The Forge build that SpongeForge is built against and is guaranteed to be compatible with". Describe what this means outside of the table.

source/versions/filenames.rst
```diff
@@ -0,0 +1,117 @@
...
+| ``LatestAPI``        | The minor version of SpongeAPI implemented by this file (the ``Y`` in                         |
+| ``MinorRelease``     | :ref:`semantic versioning<sem-ver>`).                                                         |
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``Recommended``      | The released version of the implementation when **not** followed by ``-RC<BuildNumber>.``     |
+| ``Version``          |                                                                                               |
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``-RC``              | The build number in development for the next recommended release. When a build number is      |
+| ``<BuildNumber>``    | present, the ``RecommendedVersion`` has **not** been released yet.                            |
++----------------------+-----------------------------------------------------------------------------------------------+
+
```
dualspiral, line 39: I would move the description of the significance of the forge build number here. You can then say:

* SpongeForge will work with the Forge build ``ForgeBuild``
* SF may work with other Forge versions, but we don't guarantee this
* We **tend** to compile against the latest Recommended Build for Forge

The note, warning and description are too far away from one another, bring them all together here (though it was an artifact of what you started with!). Show what the version string is, then start talking about its significance, not the other way around.

source/versions/legacy-versions.rst
```diff
@@ -0,0 +1,37 @@
+=============================
+Older Versions of SpongeForge
+=============================
+
+.. note::
+
+    Our versioning policy was updated in October 2018. **Prior to** SpongeAPI 8.0, the following information applies.
```
dualspiral: 8.0 -> 7.1

source/versions/filenames.rst
dualspiral: I agree with @ImMorpheus (with my caveats, above), [but I just want to point out this was a copy/paste](https://github.com/SpongePowered/SpongeDocs/pull/785/files/3925d9afa55f368d4d6300331e443d9da05e4f19#diff-1b21b02d7e6287095c0a769eab09ebc1L64). This is a repeat of what's already been said anyway, so this entire warning could probably be removed.

dualspiral (commented):
I'm not convinced about splitting into versions and filenames, as the filename for our impls is pretty much the version. I would perhaps split it into API and Implementation versions and rejig stuff so server owners and plugin developers have targetted pages. It doesn't invalidate anything that's happened here as most of what is written may just need a cut and paste job into different homes.

So, something like this:

```
                                  |--> SpongeAPI Versioning & Filenames
Versioning (inc. common stuff) ---|--> Legacy SF/V Versioning
                                  |--> SpongeForge/Vanilla Versioning & Filenames
```

I would hold off on doing this until that's discussed though - **this isn't a request for change but a request for input** to try to ensure that duplication is minimised and that people can minimise the pages they look at (the quick start guide for setting up a server came about because we had a lot of pages for how to set up a server, but no "reference page").

It would be instructive, should there be any confusion, to read https://github.com/SpongePowered/SpongeForge/issues/2367 where the version changes were proposed which might give clarity.

Else, I have a few comments below.

source/versions/filenames.rst
```diff
@@ -0,0 +1,117 @@
...
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``Time``             | The time when the build job ran.                                                              |
++----------------------+-----------------------------------------------------------------------------------------------+
+| ``<BuildNumber>``    | The build number of Sponge. Supply this number when reporting bugs or seeking support.        |
++----------------------+-----------------------------------------------------------------------------------------------+
+
+.. note::
+
+    The first format without the build information is the release and recommended build format. The second format with 
+    build information is the *SNAPSHOT* format. This version is still in development.
```
gabizou: Maybe mention that this format is only used as raw downloads. Depending on the API via Gradle or Maven will download a formatted named jar as the recommended build format (`spongeapi-7.2.0-SNAPSHOT.jar`). Downloading via browser for specific builds will result in the timestamped jar file.

source/versions/filenames.rst
Grauldon: Done

source/versions/filenames.rst
Grauldon: Warning removed; note handled in accordance with [comments below](https://github.com/SpongePowered/SpongeDocs/pull/785#discussion_r261731618).

source/versions/filenames.rst
Grauldon: Corrected

source/versions/filenames.rst
Grauldon: Changed

source/versions/filenames.rst
Grauldon: Removed warning, moved the note to below table and described Forge build between the table and the note. Copied changes to the legacy-versions.rst file (except I excluded "**tend** to" in statement about Forge's Recommended Build).

source/versions/legacy-versions.rst
Grauldon: Corrected

source/versions/filenames.rst
Grauldon: `spongeapi-7.2.0-SNAPSHOT.jar` is the recommended build format? I thought the `-SNAPSHOT` label means it is still in development.

source/versions/versioning.rst
Grauldon: Updated

source/versions/versioning.rst
Grauldon: Changed

source/versions/filenames.rst
Grauldon: Added another format. May need to change further to have the appropriate label.

Grauldon:
I rewrote versioning.rst and corrected/improved filenames.rst.

>  I would perhaps split it into API and Implementation versions and rejig stuff so server owners and plugin developers have targetted pages.

@dualspiral I get what you are saying, and I agree, although I had already made some changes along these lines. I left the changes I had made to see if they suffice and to allow input from others. It wouldn't take much time to make the changes, so we can go that route if decided. I think your solution might be better since I am concerned that readers will brush over the warning and read/use the balance of the page anyways.

Also, I have read #2367 quite a bit; actually, I have printed sections of it and highlighted parts throughout the pages. I am concerned this may be tripping me up though. I read all comments and try to implement them with the explanations in that issue as backdrops. My concern is that some of the information may have changed, but I keep stumbling because I am not aware of those changes. I don't want to come across as not listening or being hard headed. I'm just trying to use feedback in light of the explanations in the issue.

@gabizou Please check comments concerning the SpongeAPI filename format.

Thank you all for the comments and feedback!

source/versions/filenames.rst
```diff
@@ -57,22 +63,6 @@ version ``1.12.2``, was built with Forge ``14.23.4.2705`` (Build ``2705``), guar
...
-.. note::
-
-    Normal Forge mods can usually run on any build of Forge for a given Minecraft version (e.g. 1.8.0) without any
-    problem. However, SpongeForge needs to access, among other things, internal parts of Forge, which most mods
-    shouldn’t be touching, let alone modifying as Sponge does. Since Forge is free to change their internal code
-    whenever they want to, its normal guarantee of backwards-compatibility doesn’t apply to SpongeForge. Feel free to
-    use more recent versions of Forge, than the one used for SpongeForge, but we can't always guarantee compatibility.
-
-.. warning::
-
```
phit: this still applies, why did you remove it?

source/versions/versioning.rst
```diff
@@ -4,115 +4,111 @@ Versioning
...
-    ``spongevanilla-1.12.2-7.2.0`` is a released version of the API and a released version of the implementation. It is 
-    also a recommended version because the *RC* label is not used.
+Recommended Versions are also known as Recommended Builds or Releases. They are versions of reasonable quality of which 
+the implementations can make full use of the functionality available in SpongeAPI, and plugins compatible with the API 
+are guaranteed to work as designed.
+
+.. warning::
+
+    The following information is for Sponge Developers. Plugin Developers and Server Administrators should only use 
+    **Recommended Versions** for development and server installations. Doing otherwise is at your own risk and **not**
```
phit: sounds a bit too harsh, but I guess this is fine?

source/versions/versioning.rst
```diff
@@ -4,115 +4,111 @@ Versioning
...
+
+    The ``-SNAPSHOT`` label on a SpongeCommon file means the implementation is nearly complete but is still not a 
+    stable release. More importantly, Gradle determines whether or not to append the *-RC{build number}* to the version 
+    string by the presence of the label.
+
+Release Candidate
+=================
+
+Release Candidates (**RC**) are builds in which bug fixes are released for implementations while they are still in 
+development. They are nearly ready for a full release, but require more bug fixes and testing.
```
phit: as we run rolling releases, every commit is a RC!
they can range from not even launching to just as stable as a rb

source/versions/filenames.rst
Grauldon: Oh, snap! At first, I saw it as a duplication with the new content. Had I seen the issue tracker part, I would have misunderstood a previous comment to mean not to use the issue tracker. I will put it back in.

source/versions/versioning.rst
Grauldon: How about "Using a version that is not a recommended build introduces instability and nearly guarantees problems. We strongly discourage the use of non-recommended builds by plugin developers and server administrators."?

Other options are just removing the last sentence or using dualspiral's idea of separate pages.

source/versions/versioning.rst
Grauldon: I don't understand what you're saying. Should I remove this section, add a statement about the rolling releases, or am I missing the release candidate idea completely?

phit: Yes, anything that isn't an RB is an RC, it isn't just stuff that isn't ready, it's everything that gets pushed to Github doesn't matter what state it is in. Any RC can become an RB: e.g. these 3 builds are identical from the codebase https://i.imgur.com/wkClA3j.png

source/versions/filenames.rst
Grauldon: Corrected.

source/versions/versioning.rst
Grauldon: Changed to statements above.

gabizou (approved):

source/versions/versioning.rst
```diff
@@ -104,9 +105,11 @@ A release which has the ``-SNAPSHOT`` label represents the next API version in d
...
 =================
 
-Release Candidates (**RC**) are builds in which bug fixes are released for implementations while they are still in 
-development. They are nearly ready for a full release, but require more bug fixes and testing.
-
+Any push made to GitHub which is not a *Recommended Build* is a *Release Candidate*. Any release candidate can become a 
+recommended build. Furthermore, the code may or may not change; the change might occur in a settings file. The 
+resulting build may run as well as a recommended build, or it may not run at all. In the end, a change occurred in the 
+repository, it triggered a build, and the resulting build is not a recommended build. Therefore, it is a release 
+candidate.
```
Inscrutable: Brick-a-budgie, that's horrible to read. Accurate but rambling, methinks it has too many words by far. I mean, the only thing we really need to get across is that RC's *might not work properly*, but may become RB's after further testing. Getting into the nuts-and-bolts of the process (e.g. "triggered a build") is unnecessary, imho.

Grauldon:
Added two more commits:

1. Found another case where a relative path was more than two deep
2. Found changes that should have occurred with the original movement of the filename explanation

Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/.

- added: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/versions/filenames.html
- added: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/versions/index.html
- added: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/versions/legacy-versions.html
- added: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/versions/versioning.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/contributing/versioning.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/index.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/plugin/api-versions.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/plugin/internals/implementation.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/server/getting-started/implementations/spongeforge.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/server/getting-started/implementations/spongevanilla.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/server/spongineer/debugging.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b9c4ad9/server/spongineer/logs.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/785#issuecomment-471165504), not to the direct url to the preview._

source/plugin/api-versions.rst
```diff
@@ -37,41 +40,7 @@ Getting the API Version from Implementations
...
-    you should use plugins built against 7.1.0 unless specifically testing new functionality as asked by plugin devs, or
-    you are willing to put up with small breakages occasionally for the sake of new features.
-
-    This **WILL** be fixed in the near future. When this happens the Version number of SpongeForge & SpongeVanilla will
-    no longer reference unreleased SpongeAPI minor versions, but will still ship previews of the next minor API for
-    testing purposes.
-
-    Major API version bumps will be reserved for the experimental / bleeding builds, and will be very explicit as their
-    major version number will be different.
+See :doc:`../../versions/versioning`.
```
ST-DDT: Use absolute paths here.

source/plugin/internals/implementation.rst
```diff
@@ -18,7 +18,7 @@ SpongeCommon
 
 - **Group ID**: ``org.spongepowered``
 - **Artifact ID**: ``spongecommon``
-- **Version**: Same as SpongeAPI
+- **Version**: Same as SpongeAPI + current recommended version (see :doc:`../../../versions/versioning` for details)
```
ST-DDT: and here

source/server/spongineer/debugging.rst
```diff
@@ -23,7 +23,7 @@ Is the recommended Forge version installed?
 Usually SpongeForge will run on older or newer Forge builds than the recommended build.
 However, it is strongly advised to run the recommended build only.
 If you encounter a crash and your versions are mismatching, match them first and try again.
-If you're unsure which Forge build you need, take a look at :doc:`/server/getting-started/implementations/spongeforge`
+If you're unsure which Forge build you need, take a look at the **SpongeForge** section of :doc:`../../versions/filenames`.
```
ST-DDT: and here

source/server/spongineer/debugging.rst
```diff
@@ -95,7 +95,7 @@ Mismatched SpongeForge and Forge
 
 This is a common crash when you try to run SpongeForge on the wrong Forge build. Note that the target/Mixin can vary.
 Always match Forge against SpongeForge! If you're unsure which version of Forge is required and you already got your
-SpongeForge build, take a look at: :doc:`../getting-started/implementations/spongeforge/`
+SpongeForge build, take a look at the **SpongeForge** section of :doc:`../../versions/filenames`.
```
ST-DDT: and here

source/server/spongineer/logs.rst
```diff
@@ -182,7 +182,7 @@ The example log indicates that we're running:
  [main/INFO] [FML]: Loading tweaker org.spongepowered.asm.launch.MixinTweaker from spongeforge-1.12.2-2705-7.1.0-BETA-3136.jar
 
 This indicates that SpongeForge 3136 was found and loaded by Forge. For further help regarding the SpongeForge
-naming scheme, have a look here: :doc:`../getting-started/implementations/spongeforge/`.
+naming scheme, have a look at :doc:`../../versions/filenames`.
```
ST-DDT: and here

ST-DDT (commented):
Nice changes!

source/plugin/api-versions.rst
Grauldon: Changed

source/plugin/internals/implementation.rst
Grauldon: Changed

source/server/spongineer/debugging.rst
Grauldon: Changed

source/server/spongineer/debugging.rst
Grauldon: Changed

source/server/spongineer/logs.rst
Grauldon: Changed

Inscrutable:
@ImMorpheus - your review is the last hurdle. If you're happy with the changes we can publish this.

source/plugin/internals/implementation.rst
```diff
@@ -46,7 +46,7 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongevanilla:1.12.2-7.1.0-BETA-116:dev'
+        compile 'org.spongepowered:spongevanilla:1.12.2-7.1.0-RC116:dev'
```
ImMorpheus: This won't compile.

The version change happened with RC123. 
There's no spongevanilla:1.12.2-7.1.0-RC116.

See https://www.spongepowered.org/downloads/spongevanilla/stable/1.12.2

source/plugin/internals/implementation.rst
```diff
@@ -65,5 +65,5 @@ Example Using Gradle
 .. code-block:: groovy
 
     dependencies {
-        compile 'org.spongepowered:spongeforge:1.12.2-2705-7.1.0-BETA-3442:dev'
+        compile 'org.spongepowered:spongeforge:1.12.2-2705-7.1.0-RC3442:dev'
```
ImMorpheus: Same as above.

The versioning change for SF happened with RC3484

source/versions/filenames.rst
```diff
@@ -0,0 +1,126 @@
...
+Examples
+--------
+
+SpongeForge Jar files will always follow this naming scheme to allow you to easily identify compatibility.
+
+For example, the file name ``spongeforge-1.12.2-2705-7.1.4.jar`` is compatible with Minecraft version ``1.12.2``, was 
+built with Forge ``14.23.4.2705`` (Build ``2705``), guarantees compatibility with SpongeAPI ``7.1.0``, is a recommended 
+version, and is the ``4`` th release of SpongeForge with this API.
+
+Another example is the file name ``spongeforge-1.12.2-2705-7.1.5-RC3449.jar``. This file is compatible with Minecraft
```
ImMorpheus: There's no spongeforge-1.12.2-2705-7.1.5-RC3449, but this is used as an example so I'm not sure it's worth changing

source/versions/legacy-versions.rst
```diff
@@ -0,0 +1,51 @@
...
+    version mismatch.
+    Even though there will be no guarantee of compatibility, please report any breakage to the issue tracker, so that
+    SpongeForge can be updated.
+
+Example
+-------
+
+SpongeForge Jar files will always follow this naming scheme, to allow you to easily identify compatibility.
+
+For example, the file name ``spongeforge-1.12.2-2705-7.1.0-BETA-3442.jar`` is compatible with Minecraft version
```
ImMorpheus, line 49: See above

ImMorpheus (changes requested):

source/plugin/internals/implementation.rst
Grauldon: Corrected

source/plugin/internals/implementation.rst
Grauldon: Corrected

source/versions/filenames.rst
Grauldon: I agree with using correct filenames, so I went ahead and corrected them. I also updated the SpongeVanilla example as well.

source/versions/legacy-versions.rst
Grauldon: This filename was released (see [SpongeForge download](https://www.spongepowered.org/downloads/spongeforge/stable/1.12.2?until=2018-09-14T15%3A28%3A16Z)), so I left it.

ImMorpheus (approved):

source/versions/filenames.rst
```diff
@@ -0,0 +1,126 @@
...
+.. _sponge-forge-file-name:
+
+SpongeForge
+===========
+
+The format of the SpongeForge download filename is:
+
+``spongeforge-<MCVersion>-<ForgeBuild>-<APIMajor>.<LatestAPIMinorRelease>.<RecommendedVersion(-RC<BuildNumber>)>.jar``
+
++----------------------+-----------------------------------------------------------------------------------------------+
```
ST-DDT, line 22: I'm not sure, but the tables without headers look a little bit strange. Should we add the headers?

ST-DDT (approved):
I'm unsure about the missing table headers, but otherwise it looks good to me.
Preview updated.

dualspiral (approved):
I'll add my review! :) 

I think this is good to go. We can fix minor quibbles later as they come up, no point batting this around further.

phit:
should versioning be a main category? if it should it should be moved down in the sidebar

ST-DDT:
> should versioning be a main category? 

Yes, unless you have a good idea where to place it otherwise.

> it should be moved down in the sidebar

I somewhat agree with that. Any suggestion for the target ordering?

phit:
probably like this, so after creating a server and before dev work as kinda a bridge

    Creating a Server
    Versioning Policy
    Preparing for Development
    Creating a Plugin
    Ore Documentation
    Contributing to Sponge
    About the Sponge Project

the ordering and naming kinda needs a general makeover anyway but thats out of scope for this pr

Grauldon:
I like phit's suggestion on the sidebar. I can also add the table headers while making that update, perhaps `Component` for the first column and `Purpose` for the second. 

I also realized I forgot a parenthetical note on `Recommended Version` for the SpongeForge table while looking at the latest comments. I want to update it from 

> The released version of the implementation when **not** followed by ``-RC.``

to

> The released version of the implementation when **not** followed by ``-RC`` (the ``Z`` in :ref:`semantic versioning`). 

It will look just like the parenthetical notes in ``APIMajor`` and ``LatestAPIMinorRelease`` (except for stating Z), and it will complement Sponge's use of SemVer.

Any thoughts on table headers and the note or further comments about phit's suggestion?

Inscrutable:
Those headless tables seem OK by me, they are just another way of doing a list. 
I agree with dualspiral. 🏏 Add the note, tweak the order, and it can be finished.

PR #787, opened 2019-02-11
Add quick start guide for creating a server

dualspiral:
From some conversation some time ago, it was pointed out that there was no "quick start" guide for creating a server, the steps were split over several pages with lots of information that could seem daunting. So, I created such a page.

The idea behind this document is we can point people to this which is a less daunting page than multiple documents on how to set up their servers, while it contains links to relevant places in the docs if they need more help.

It's in a good state, but I'm happy to consider this WIP. Docs staff: feel free to just commit on top of this if you wish to do so.

### DIFF
--- a/source/server/quickstart.rst
+++ b/source/server/quickstart.rst
@@ -1,0 +1,147 @@
+=================
+Quick Start Guide
+=================
+
+This is a step by step reference on how to create a Minecraft server with Sponge.
+
+.. note::
+
+    Are you migrating from a Bukkit or Canary based server? Please read
+    :doc:`getting-started/migrating` as it explains the differences between these platforms and Sponge,
+    and explains the migration process.
+
+.. warning::
+
+    If you're running behind a server proxy such as BungeeCord, Waterfall or Velocity, there are additional
+    steps you must follow once Sponge is installed, which are documented in
+    :doc:`getting-started/bungeecord`.
+
+System Requirements
+===================
+
+Sponge requires Java 8 update 20 or higher to run. Sponge **does not** work with Java 9 or above.
+We also recommend a CPU with a large single core speed and at least 2GB of RAM (higher if you
+plan to run with mods). We highly recommend using the 64bit Java runtime.
+
+See :doc:`getting-started/jre` for more help with how to install Java 8.
+
+SpongeForge or SpongeVanilla?
+=============================
+
+There are two official Sponge implementations on offer that integrate directly with the official
+Minecraft server jar:
+
+* **SpongeForge** - a mod for Forge that allows you to use Sponge plugins with Forge mods;
+* **SpongeVanilla** - a server jar that adds Sponge support directly into vanilla Minecraft.
+
+When choosing, there are a few things to keep in mind:
+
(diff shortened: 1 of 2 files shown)


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b6ca406/.

- added: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b6ca406/server/quickstart.html
- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b6ca406/server/index.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/787#issuecomment-462501351), not to the direct url to the preview._

source/server/quickstart.rst
```diff
@@ -0,0 +1,147 @@
...
+* :doc:`spongineer/commands`
+
+.. _quickstart-when-things-go-wrong:
+
+When Things Go Wrong
+====================
+
+From time to time, incompatible mods or other strange behavior might cause problems on your server. Before
+reaching out for help, following the advice on these pages may help you fix your problems quickly:
+
```
dualspiral, line 144: Question, should we include our `oldmixin` and `foamfix` messages in Discord here (or above)? They are common errors so would they benefit from being part of such a guide?

Inscrutable: Including it would be helpful, but it might warrant a separate page, if there's enough content to go in it. Such advice is mod specific in some cases, so it may be more volatile than our usual content.

source/server/quickstart.rst
```diff
@@ -0,0 +1,147 @@
...
+Quick Start Guide
+=================
+
+This is a step by step reference on how to create a Minecraft server with Sponge.
+
+.. note::
+
+    Are you migrating from a Bukkit or Canary based server? Please read
+    :doc:`getting-started/migrating` as it explains the differences between these platforms and Sponge,
+    and explains the migration process.
```
Inscrutable, line 11: I think the second "explains" is redundant.

Inscrutable (approved):
I was horrified to discover I could only find one thing to quibble about. Nice work!

source/server/quickstart.rst
dualspiral: I think you're right.

dualspiral: Ah, hold on, I know why I put that. It's because of the comma before the last "and" to separate it from the "these platforms and Sponge" section. The statement would become (removing the comma as it makes less sense at this point)...

> as it explains the differences between these platforms and Sponge and the migration process

which would infer that I'm describing the differences between all three. Maybe changing "explains" to "sets out" or something similar will make it different enough to state that there are two points being made, rather than one amalgamated one?

> as it explains differences between these platforms and Sponge, and sets out the migration process

May also just need a complete reword.

Inscrutable: If you leave the comma in it seems less ambiguous, but it's not at all necessary to change it anyway. Better to merge it than make word-salad.

PR #790, opened 2019-02-16
Many many staff changes.

Inscrutable:
There's another step I want to do here, which is part of the reason this is a PR instead of a direct edit. 
The current staff list is more suited to bragging rights than actually helping outsiders identify our staff. This is because each person is listed only once, with all their roles. This was done to avoid duplication, but it makes it harder for someone to, for example, identify all the Ore Reviewers, as they are scattered around the page. This inconveniences users and our own staff sometimes.

I propose to expand it out with complete lists for each role (again) so that it is easier to identity all the staff in a given role, even if that does mean kashike ends up appearing everywhere. Comments are welcome.


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/64fa690/.

- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/64fa690/about/staff.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/790#issuecomment-464276871), not to the direct url to the preview._

ST-DDT (approved):

Grauldon:
> it makes it harder for someone to, for example, identify all the Ore Reviewers

I agree

> I propose to expand it out with complete lists for each role (again)

Are you suggesting adding below the current page a list of roles with each staff member assigned that role?

I suggest leaving the page as it is, but adding members to the different roles. For example, kashike is the Discord Manager, a Developer, Ore Reviewer, and System Administrator. I suggest leaving kashike name in the Project Managers table and adding it to the tables under API and Implementation Developers, Ore Reviewers, and System Administrators. Further, continue to list out each role for each person, so kashike will have all four listed in all four tables.

Having all the roles listed in each table allows for easy and quick reading. I may be looking for those who are Ore Reviewers, but I will quickly see that kashike is also a developer, system admin, and head of Discord. If I am looking for developers, I can see kashike is also an Ore Reviewer (plus the other roles).

Inscrutable:
To clarify; I want to have definitive lists for each staff category, _replacing_ the current hybrid model. I don't particularly want that to bloat it more by having everyone's roles listed everywhere they appear. It's a lot harder to find a use case for "what are the roles of this person" than "who are the reviewers". The former is mostly internal use, and we have a secret table to manage that.

PR #794, opened 2019-03-20
Update to config list + tool

12awsomeman34:
Sorry this took wayyy too long to get around to, but here it is.

I'm not sure what version of the config list we want to maintain (latest stable (e.g. 7.1.5), latest release (7.1), or just w/e) but I've updated it here to show the tool.

It's a plugin that creates a file in the root directory of the server (`config-dictionary.txt`) from there you copy + paste into the docs conf file. It does this by fetching child nodes recursively to map out and format the config list. It wouldn't be difficult if we decide to expand the list to include more info like the table did (default value, value type, etc.) in the future.

### DIFF
--- a/source/server/getting-started/configuration/sponge-conf.rst
+++ b/source/server/getting-started/configuration/sponge-conf.rst
@@ -525,7 +525,7 @@ world-enabled
 
 ------------------------------------------------------------------------------------------------------------
 
-This config was generated using SpongeForge build 3442 (with Forge 2768), SpongeAPI version 7.1.5:
+This config was generated using SpongeForge recommendation 5 (SpongeForge 3554 with Forge 2768), SpongeAPI version 7.1:
 
 .. code-block:: guess
 
@@ -544,7 +544,7 @@ This config was generated using SpongeForge build 3442 (with Forge 2768), Sponge
             # A list of mod ids that have broken network handlers (they interact with the game from a Netty handler thread).
             # All network handlers from a forcibly scheduled to run on the main thread.
             # Note that this setting should be considered a last resort, and should only be used as a stopgap measure while waiting for a mod to properly fix the issue.
-            broken-network-handler-mods=null
+            broken-network-handler-mods=[]
         }
         bungeecord {
             # If 'true', allows BungeeCord to forward IP address, UUID, and Game Profile to this server.
@@ -910,7 +910,7 @@ This config was generated using SpongeForge build 3442 (with Forge 2768), Sponge
         }
         player-block-tracker {
             # Block IDs that will be blacklisted for player block placement tracking.
-            block-blacklist=null
+            block-blacklist=[]
             # If 'true', adds player tracking support for block positions.
             # Note: This should only be disabled if you do not care who caused a block to change.
             enabled=true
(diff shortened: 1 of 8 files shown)


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/030c524/.

- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/030c524/server/getting-started/configuration/sponge-conf.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/794#issuecomment-475034229), not to the direct url to the preview._

Inscrutable:
Thanks! Where possible, we should target the latest recommended build. 
For the moment, the config text is published as a courtesy so translations can happen. If we add any description or text in addition to the config, it will be harder to maintain. Additional data might need to be put into a separate section of the page, or perhaps as notes or tips to flag them. Values and examples could be useful, as the default formatting still defeats some users.

ST-DDT:
@NariKiro I updated to the lastest SpongeForge recommended build for you.

source/server/getting-started/configuration/sponge-conf.rst
```diff
@@ -470,7 +525,7 @@ world-enabled
 
 ------------------------------------------------------------------------------------------------------------
 
-This config was generated using SpongeForge build 3442 (with Forge 2705), SpongeAPI version 7.1.0:
+This config was generated using SpongeForge recommendation 5 (SpongeForge 3554 with Forge 2768), SpongeAPI version 7.1:
```
ST-DDT, line 528: Is there a better way to refer to the recommended build 7.1.5?

ST-DDT (changes requested):
We also need to check the entity collisions block (Nested blocks).
See also: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/ae0deaf/server/getting-started/configuration/sponge-conf.html#entity-collisions / https://github.com/SpongePowered/SpongeDocs/pull/794/files#diff-a403cb94bdc45d307e21dee1ac3ce461R145

ST-DDT:
As for the nested configuration options should we add more headline levels there?

@NariKiro Do you want to give it a try?

source/server/getting-started/configuration/sponge-conf.rst
Inscrutable: That *should* be `SpongeForge recommended build 1.12.2-2768-7.1.5`, (which is equivalent to `release candidate 12.2-2768-7.1.5-RC3554`).

12awsomeman34: Uh, I don't actually remember changing this part. I'll change it to 7.1.5.
Edit: nvm just looked at commit history

12awsomeman34:
Yeah I can look into that.

ST-DDT:
I found [this](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings) page that explains the headline levels. But it looks like you can omit levels and it will adjust the other levels accordingly.

ST-DDT:
I started looking into this one yesterday again and completed the necessary changes.
However, this made me realize some limitations of the current design/my suggestions for it.

The deep nesting structure makes is practically impossible to differentiate the headline levels on the lower levels (5% difference in size). So we might have to switch to a type reference link based approach. The final result will then look somewhat like:

> ## Global Config
>
> * **enabled**
>     Description: Lorep ipsum
>     Type: `boolean`
>     Default: `true`
> * **world**
>     Description: Lorep ipsum
>     Type: [World Config](#world-config)
> * **mods**
>     Description: Lorep ipsum
>     Type: [Mods Config](#mods-config)
>
> ## World Config
> ...

For that purpose I will switch over to use the actual config classes from SpongeCommon, which allows me to analyze the field types. I will do so in a different PR.

Any thoughts or ideas regarding this?

Inscrutable:
Is there something worth merging here, or should this wither and (hopefully) be replaced by a new PR?
The config page is largely useless for translation purposes without it.

PR #795, opened 2019-04-16
Add AdoptOpenJDK link

ghost:
I don't like that I need an Oracle account just to download the JDK. Here I suggest the AdoptOpenJDK alternative.

### DIFF
--- a/source/preparing/jdk.rst
+++ b/source/preparing/jdk.rst
@@ -21,6 +21,6 @@ Oracle provides free downloads of the Java Development Kit on their website. Ens
 
 * `Java Development Kit 8 <https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html>`__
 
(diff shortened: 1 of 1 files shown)


worm424 (approved):
Talked with Duel looks all good to me.

JBYoshi (approved):
Thanks for this!

source/preparing/jdk.rst
```diff
@@ -21,4 +21,6 @@ Oracle provides free downloads of the Java Development Kit on their website. Ens
 
 * `Java Development Kit 8 <https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html>`__
 
+If you don't want to create an Oracle account in order to download, you can also download the JDK from `AdoptOpenJDK <https://adoptopenjdk.net/?variant=openjdk8&jvmVariant=hotspot>`__.
```
Inscrutable: I think it's a bit misleading as-is. The AdoptOpenJDK is not the same beast as the Oracle JDK, as far as I know, and I'm not sure that's the right way to say "an Oracle account is required to download Oracle's JDK". Perhaps instead: 
"An open-source alternative is the OpenJDK, which is available from (yadda yadda)"

ST-DDT (approved):

PR #801, opened 2019-05-17
Added section for merging asset nodes

ghost:
Created an additional section in the AssetAPI docs for how to merge nodes from an existing Asset to a current configuration node. Have had several people ask for something like this and it seems to be better to set this here than in the configuration node section as chronologically the user may not have looked at assets yet which is required here.

### DIFF
--- a/source/plugin/assets.rst
+++ b/source/plugin/assets.rst
@@ -59,16 +59,3 @@ following code:
     ``getResourceAsStream`` methods in ``Class``\es and ``ClassLoader``\s. **These methods should generally be avoided**
     within the SpongeAPI environment in favor of the ``AssetManager`` in order to provide a more confluent way of
     retrieving resources not only within your own plugin, but for other plugins as well.
-
-Working with Assets
-===================
-
-If you would like to merge new nodes and their values to your existing configuration file you can use your
-``CommentedConfigurationNode`` and load values from a given asset from inside your resources.
-
-.. code-block:: java
-
-    node.mergeValuesFrom(HoconConfigurationLoader.builder()
-                        .setURL(plugin.getAsset("default.conf").get().getUrl())
-                        .build()
-                        .load(ConfigurationOptions.defaults()));
--- a/source/plugin/assets.rst
+++ b/source/plugin/assets.rst
@@ -59,3 +59,16 @@ following code:
     ``getResourceAsStream`` methods in ``Class``\es and ``ClassLoader``\s. **These methods should generally be avoided**
     within the SpongeAPI environment in favor of the ``AssetManager`` in order to provide a more confluent way of
     retrieving resources not only within your own plugin, but for other plugins as well.
+
+Working with Assets
+===================
+
+If you would like to merge new nodes and their values to your existing configuration file you can use your
+``CommentedConfigurationNode`` and load values from a given asset from inside your resources.
+
+.. code-block:: java
+
+    node.mergeValuesFrom(HoconConfigurationLoader.builder()
+                        .setURL(plugin.getAsset("default.conf").get().getUrl())
+                        .build()
+                        .load(ConfigurationOptions.defaults()));
--- a/source/plugin/configuration/loaders.rst
+++ b/source/plugin/configuration/loaders.rst
@@ -166,7 +166,7 @@ If you have an extra configuration class, you can use a much easier approach tha
 config is missing. See also the examples on the :doc:`serialization` page.
 
 Merging Nodes from an Asset
-~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If you would like to merge new nodes and their values to your existing configuration file you can use your
 ``CommentedConfigurationNode`` and load values from a given asset explained above. This will take each node in 
--- a/source/plugin/configuration/loaders.rst
+++ b/source/plugin/configuration/loaders.rst
@@ -12,6 +12,7 @@ Configuration Loaders
     ninja.leaping.configurate.loader.AbstractConfigurationLoader.Builder
     ninja.leaping.configurate.loader.ConfigurationLoader
     org.spongepowered.api.asset.AssetManager
+    org.spongepowered.api.asset.Asset
     java.lang.String
     java.net.URL
     java.nio.file.Path
@@ -135,8 +136,10 @@ Again, errors will be propagated as an ``IOException`` and must be handled.
     strongly recommended to do this outside of the main thread. See also common :doc:`/plugin/practices/bad` you should
     avoid.
 
-Example: Loading a Default Config from the Plugin Jar File
+Loading a Default Config from the Plugin Jar File
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A popular way to provide a default configuration file with your plugin is to include a copy of it in your plugin jar, copying it to the config directory when the config file has yet to be created. You can use :doc:`the Asset API page <../assets>` to do this as shown in the example below:
 
 .. code-block:: java
 
@@ -146,13 +149,9 @@ Example: Loading a Default Config from the Plugin Jar File
 
 For this example it is important to note that the :javadoc:`AssetManager#getAsset(String)` method works relative to the
 plugin's asset folder. So, if in the above example the plugin ID is ``myplugin``, the ``default.conf`` file
-must not lie in the jar file root, but instead in the directory ``assets/myplugin``. For more information, see
-:doc:`the Asset API page <../assets>`.
-
-This variant should be used instead of checking whether the root node is empty as it simplifies the process.
-The first argument is the path in which to set the node, the second represents whether or not 
-the file should be overriden each time, and the final argument represents if the existing 
-config should be overriden only if absent.
+must not lie in the jar file root, but instead in the directory ``resources/assets/myplugin``. This example also uses 
+:javadoc:`Asset#copyToFile(String, boolean, boolean)` which allows an the file creation to override existing
+files only if specified. 
 
 .. note::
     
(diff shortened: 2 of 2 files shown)


Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/3b0ac52/.

- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/3b0ac52/plugin/configuration/loaders.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/801#issuecomment-493345607), not to the direct url to the preview._

source/plugin/assets.rst
```diff
@@ -59,3 +59,16 @@ following code:
     ``getResourceAsStream`` methods in ``Class``\es and ``ClassLoader``\s. **These methods should generally be avoided**
     within the SpongeAPI environment in favor of the ``AssetManager`` in order to provide a more confluent way of
     retrieving resources not only within your own plugin, but for other plugins as well.
+
+Working with Assets
```
ST-DDT: This headline exists already.

ST-DDT (changes requested):
IMO this section should be better moved to the config section. From the context it does not match here.

Maybe update this page:
https://docs.spongepowered.org/stable/en/plugin/configuration/loaders.html?highlight=config#example-loading-a-default-config-from-the-plugin-jar-file

source/plugin/assets.rst
ghost: > IMO this section should be better moved to the config section. From the context it does not match here.
> 
> Maybe update this page:
> https://docs.spongepowered.org/stable/en/plugin/configuration/loaders.html?highlight=config#example-loading-a-default-config-from-the-plugin-jar-file

Yeah I can see It making sense that way. Put it in assets originally as assets explanation came after but that link shows assets are linked prior and an example is shown so they dev should understand by then. Newest commit will reflect those changes. And fixed that header issue; thought I had changed that earlier honestly

ST-DDT:
The two sections at the end of the page look redundant, don't they? Can you merge them together? The old code example has been argued against in the past so you could replace it with yours. Or explain the differences that both yield.

ghost:
> The two sections at the end of the page look redundant, don't they? Can you merge them together? The old code example has been argued against in the past so you could replace it with yours. Or explain the differences that both yield.

Would say the default config section for loaders is a bit redundant considering its explicitly referring to using an asset. That and there is a good bit of unneeded code there as that entire segment could be done in one line via `Asset#copyToFile(asset, false, true). Might have to edit a few things

source/plugin/configuration/loaders.rst
```diff
@@ -26,7 +26,7 @@ configurations, giving you the option of hard-coding default values or loading f
     The default and recommended config format for Sponge plugins is HOCON.
 
 Getting your Loader
-~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
ST-DDT: Unneccesary change

source/plugin/configuration/loaders.rst
```diff
@@ -140,22 +140,17 @@ Example: Loading a Default Config from the Plugin Jar File
...
-                HoconConfigurationLoader.builder().setURL(jarConfigFile).build();
-        rootNode = defaultLoader.load();
-    }
 
 For this example it is important to note that the :javadoc:`AssetManager#getAsset(String)` method works relative to the
 plugin's asset folder. So, if in the above example the plugin ID is ``myplugin``, the ``defaultConfig.conf`` file
 must not lie in the jar file root, but instead in the directory ``assets/myplugin``. For more information, see
 :doc:`the Asset API page <../assets>`.
 
+Additionally, instead of checking to see if the node has an empty children map, this method uses ``copyToFile(path, false, true)``. The first argument is the path in which to set the node to, the second represents whether or not the file should be overriden each time, and the final argument represents if the existing config should be overriden only if absent.
```
ST-DDT: Not sure about `additionally` from the context there . There is no check for an empty node here. Maybe rephrase it somewhat like: This variant should be used instead of checking whether the root node is empty.

Please keep the line length limit in mind.

source/plugin/configuration/loaders.rst
```diff
@@ -164,3 +159,22 @@ must not lie in the jar file root, but instead in the directory ``assets/myplugi
 
 If you have an extra configuration class, you can use a much easier approach that also works if the only a part of your
 config is missing. See also the examples on the :doc:`serialization` page.
+
+Merging Nodes from an Asset
```
ST-DDT: Maybe add the `Example:`  prefix?

source/plugin/configuration/loaders.rst
```diff
@@ -164,3 +159,22 @@ must not lie in the jar file root, but instead in the directory ``assets/myplugi
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
ST-DDT: The headline marker should be as long as the headline.

source/plugin/configuration/loaders.rst
```diff
@@ -140,22 +140,17 @@ Example: Loading a Default Config from the Plugin Jar File
 
 .. code-block:: java
 
-    import java.net.URL;
-
+    Sponge.getAssetManager().getAsset(myplugin, "defaultConfig.conf").get().copyToFile(path, false, true);
```
ST-DDT: Please use the same asset name everywhere. It does not matter which one you use.

ST-DDT (changes requested):
This is way better.

ST-DDT (approved):
Looks good to me. 

It would be nice to get a short review from the API team as well.

source/plugin/configuration/loaders.rst
```diff
@@ -140,22 +140,20 @@ Example: Loading a Default Config from the Plugin Jar File
...
-    }
 
 For this example it is important to note that the :javadoc:`AssetManager#getAsset(String)` method works relative to the
-plugin's asset folder. So, if in the above example the plugin ID is ``myplugin``, the ``defaultConfig.conf`` file
+plugin's asset folder. So, if in the above example the plugin ID is ``myplugin``, the ``default.conf`` file
 must not lie in the jar file root, but instead in the directory ``assets/myplugin``. For more information, see
 :doc:`the Asset API page <../assets>`.
 
+This variant should be used instead of checking whether the root node is empty as it simplifies the process.
+The first argument is the path in which to set the node, the second represents whether or not
```
dualspiral: The first argument of what? This should be explicit in naming the `copyToFile` method - perhaps linking to it in the JDs and just explaining that the form will only copy the file to the `Path` if the file doesn't exist already.

source/plugin/configuration/loaders.rst
```diff
@@ -140,22 +140,20 @@ Example: Loading a Default Config from the Plugin Jar File
```
dualspiral, line 143: I think some of the explanation of what this example does should go here - set up the context of the example then give the example. The example and code seem to just come out of nowhere.

Actually, I don't think this should be given an "example" moniker - perhaps change the title to "Creating a Default Configuration File".

> A popular way to provide a default configuration file with your plugin is to include a copy of it in your plugin jar, copying it to the config directory when the config file has yet to be created. You can use the Asset API (link?) to do this, as shown in the example below:

source/plugin/configuration/loaders.rst
```diff
@@ -140,22 +140,20 @@ Example: Loading a Default Config from the Plugin Jar File
...
-        rootNode = defaultLoader.load();
-    }
 
 For this example it is important to note that the :javadoc:`AssetManager#getAsset(String)` method works relative to the
-plugin's asset folder. So, if in the above example the plugin ID is ``myplugin``, the ``defaultConfig.conf`` file
+plugin's asset folder. So, if in the above example the plugin ID is ``myplugin``, the ``default.conf`` file
 must not lie in the jar file root, but instead in the directory ``assets/myplugin``. For more information, see
 :doc:`the Asset API page <../assets>`.
 
+This variant should be used instead of checking whether the root node is empty as it simplifies the process.
```
dualspiral: I think you should just drop this sentence - it adds nothing.

source/plugin/configuration/loaders.rst
```diff
@@ -164,3 +162,23 @@ must not lie in the jar file root, but instead in the directory ``assets/myplugi
 
 If you have an extra configuration class, you can use a much easier approach that also works if the only a part of your
 config is missing. See also the examples on the :doc:`serialization` page.
+
+Example: Merging Nodes from an Asset
```
dualspiral: I'm probably going to receive the ire of @ST-DDT here... but this isn't an example, this is a thing you might want to do and we explain how to do them. The code is an example, perhaps, but the concept isn't and that concept is not an example of anything else displayed in the topic so far.

This is better entitled "Updating Configuration files from the Default Configuration" - or something to that effect, the title should explain the problem, "merging nodes from an asset" might just be a little obscure for some people.

dualspiral (commented):
Thanks for the submission! The biggest issue I forsee with this is the flow - it jumps around a little bit and certain parts need introducing or clarifying. I also disagree with the "example" moniker - each of these things are discrete concepts that have example code attached.

Code seems fine to me.

source/plugin/configuration/loaders.rst
ST-DDT: I'm fine with it, as long as both sections use the prefix or both don't.

source/plugin/configuration/loaders.rst
ghost: Added in next commit

source/plugin/configuration/loaders.rst
ghost: Going with prefix gone for all of them. It seems to match the flow of the page better

source/plugin/configuration/loaders.rst
```diff
@@ -163,7 +162,7 @@ config should be overriden only if absent.
 If you have an extra configuration class, you can use a much easier approach that also works if the only a part of your
 config is missing. See also the examples on the :doc:`serialization` page.
 
-Example: Merging Nodes from an Asset
+Updating Configuration Files from the Default Configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
phit: need to change underline here to match title length, that's why compile failed

ghost: Ah crap. Thought I did

PR #812, opened 2019-07-05
Fix/docs

### LINKED ISSUE #640: Outdated imports in docs

ImMorpheus:
Should also fix https://github.com/SpongePowered/SpongeDocs/issues/640

Things I left out:

- Update testplugins guidelines (https://github.com/SpongePowered/SpongeDocs/issues/793)

- the section about custom manipulators is using code from the myhomes testplugin. That plugin is broken. It should not be used as an example in its current state.

- https://github.com/SpongePowered/SpongeDocs/blob/23b8ce807cd92ec9c6026a35eeca858dba0f1e2d/source/plugin/event/custom.rst#L40-L44
is not wrong, but it's not really a good practice either. You should create an EvenContextKey and move the `Mutation` to the context.

### DIFF
--- a/source/contributing/guidelines.rst
+++ b/source/contributing/guidelines.rst
@@ -35,8 +35,8 @@ General steps
     `Minor Issues on GitHub <https://github.com/search?q=org%3ASpongePowered+Ongoing+Minor+Issue+List&type=Issues>`_.
 
 4. If the issue requires a bigger change you may want to submit the issues without the necessary changes first, so we
-   can confirm the issue and know that you're working on fixing it. You should also create a WIP (work in process) pull
-   request prefixed with ``[WIP]`` early so we can already start reviewing them.
+   can confirm the issue and know that you're working on fixing it. You should also create a draft pull
+   request or comment with ``~wip`` so we can already start reviewing them.
 
 #. Fork the project, clone it and make your changes in an extra branch.
 
--- a/source/plugin/api-versions.rst
+++ b/source/plugin/api-versions.rst
@@ -7,8 +7,8 @@ This page explains which API versions exist, and to which Minecraft version thei
 +-------------+--------------+----------------+-------------------------------------------+
 | API-Version | Release Date | End of Support | Known Implementations (Minecraft Version) |
 +=============+==============+================+===========================================+
-| *8.0.0*     | TBA          | TBA            | * *SpongeForge (1.13.x)*                  |
-|             |              |                | * *SpongeVanilla (1.13.x)*                |
+| *8.0.0*     | TBA          | TBA            | * *SpongeForge*                           |
+|             |              |                | * *SpongeVanilla*                         |
 +-------------+--------------+----------------+-------------------------------------------+
 | 7.2.0       | TBA          | TBA            | * SpongeForge (1.12.2)                    |
 |             |              |                | * SpongeVanilla (1.12.2)                  |
--- a/source/plugin/commands/commandcallable.rst
+++ b/source/plugin/commands/commandcallable.rst
@@ -82,6 +82,7 @@ passing your plugin, an instance of the command, and any needed aliases as param
     import org.spongepowered.api.command.CommandManager;
 
     CommandManager cmdService = Sponge.getCommandManager();
+    PluginContainer plugin = ...;
     cmdService.register(plugin, new MyBroadcastCommand(), "message", "broadcast");
 
 .. note::
--- a/source/plugin/commands/commandcallable.rst
+++ b/source/plugin/commands/commandcallable.rst
@@ -60,7 +60,7 @@ The first step is to create a class for the command. The class has to implement 
             return usage;
         }
 
-        public List<String> getSuggestions(CommandSource source, String arguments) throws CommandException {
+        public List<String> getSuggestions(CommandSource source, String arguments, @Nullable  Location<World> targetPosition) throws CommandException {
             return Collections.emptyList();
         }
     }
--- a/source/plugin/commands/commandcallable.rst
+++ b/source/plugin/commands/commandcallable.rst
@@ -60,7 +60,7 @@ The first step is to create a class for the command. The class has to implement 
             return usage;
         }
 
-        public List<String> getSuggestions(CommandSource source, String arguments, @Nullable  Location<World> targetPosition) throws CommandException {
+        public List<String> getSuggestions(CommandSource source, String arguments, @Nullable Location<World> targetPosition) throws CommandException {
             return Collections.emptyList();
         }
     }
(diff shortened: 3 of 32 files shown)


Inscrutable (approved):
Thanks for the clean-up. Someone else should review the code snippets. The rest seems good, especially having https everywhere. I want to comment on the "things left out":
- Move lantern link to their website 
  - it's also nice to know the project is on GitHub, we could have both.
- Update testplugins guidelines (#793)
  - may justify a separate PR
- WIP prefix for PR should be discouraged
  - please do it.
- custom manipulators using code from broken myhomes testplugin.
  - Agreed, that means either the testplugin gets fixed, a new code example is written, or we make it disappear (for now). It may justify a separate issue and PR.

source/plugin/commands/commandcallable.rst
```diff
@@ -60,7 +60,7 @@ The first step is to create a class for the command. The class has to implement
             return usage;
         }
 
-        public List<String> getSuggestions(CommandSource source, String arguments) throws CommandException {
+        public List<String> getSuggestions(CommandSource source, String arguments, @Nullable  Location<World> targetPosition) throws CommandException {
```
ST-DDT: Duplicate space after `@Nullable`

ST-DDT (approved):

source/plugin/configuration/loaders.rst
```diff
@@ -145,14 +145,17 @@ You can use :doc:`the Asset API <../assets>` to do this, as shown in the example
...
+
+    Sponge.getAssetManager().getAsset(plugin, "default.conf").get().copyToFile(path, false, true);
     loader = HoconConfigurationLoader.builder().setPath(path).build();
     rootNode = loader.load();
 
 For this example it is important to note that the :javadoc:`AssetManager#getAsset(String)` method works relative to the
 plugin's asset folder. So, if in the above example the plugin ID is ``myplugin``, the ``default.conf`` file
 must not lie in the jar file root, but instead in the directory ``assets/myplugin``. This example also uses 
-:javadoc:`Asset#copyToFile(String, boolean, boolean)` which allows an the file creation to override existing
+:javadoc:`Asset#copyToFile(Path, boolean, boolean)` which allows an the file creation to override existing
```
Grauldon: > which allows an the file creation to override existing

It's an existing error, but I think the "**an**" needs to be removed.

source/plugin/api-versions.rst
```diff
@@ -7,8 +7,8 @@ This page explains which API versions exist, and to which Minecraft version thei
 +-------------+--------------+----------------+-------------------------------------------+
 | API-Version | Release Date | End of Support | Known Implementations (Minecraft Version) |
 +=============+==============+================+===========================================+
-| *8.0.0*     | TBA          | TBA            | * *SpongeForge (1.13.x)*                  |
-|             |              |                | * *SpongeVanilla (1.13.x)*                |
+| *8.0.0*     | TBA          | TBA            | * *SpongeForge*                           |
+|             |              |                | * *SpongeVanilla*                         |
 +-------------+--------------+----------------+-------------------------------------------+
```
Grauldon, line 12: Is leaving the Minecraft version off intentional? Gabizou said to change it to 1.14, but I can agree with leaving it off for now given the amount of change and uncertainty for which Minecraft version will have fully-functional implementations.

Grauldon (changes requested):
The Minecraft version doesn't need to prevent merging, but the sentence with the extra article needs correcting while it is getting updated anyway.

Newb ALERT: I inadvertently clicked on the first "View Changes" button instead of the last one.  I hope this doesn't cause problems.

source/contributing/guidelines.rst
```diff
@@ -35,8 +35,8 @@ General steps
     `Minor Issues on GitHub <https://github.com/search?q=org%3ASpongePowered+Ongoing+Minor+Issue+List&type=Issues>`_.
 
 4. If the issue requires a bigger change you may want to submit the issues without the necessary changes first, so we
-   can confirm the issue and know that you're working on fixing it. You should also create a WIP (work in process) pull
-   request prefixed with ``[WIP]`` early so we can already start reviewing them.
+   can confirm the issue and know that you're working on fixing it. You should also create a draft pull
+   request or comment with ``~wip`` so we can already start reviewing them.
```
Grauldon, line 39: I am not certain of what is meant by
> or comment with ``~wip``

Should the comment be made in the issue or the pull request?

Perhaps one of the following two sentences would be clearer:

You should also create a draft pull request **with a** ``~wip`` **comment** so we ...

OR

You should also create a draft pull request or comment **in the issue** with ``~wip`` so we ...

source/plugin/api-versions.rst
Inscrutable: Six of one, half a dozen of the other... Leaving the MC version off an as-yet unreleased API is probably OK, as it will definitely have one by release.

source/contributing/guidelines.rst
Inscrutable: In theory WIP status can be automatically set in GitHub now. More detail is probably required.

Grauldon: I see. I found the following URLs. Maybe a link in a ``Tip`` block to one of them for more information will suffice. 

https://github.blog/2019-02-14-introducing-draft-pull-requests/

https://help.github.com/en/articles/about-pull-requests#draft-pull-requests

ImMorpheus: I was referring to limbo.
https://github.com/SpongePowered/limbo-config/wiki/Pull-Request-Commands

source/plugin/api-versions.rst
ImMorpheus: > Is leaving the Minecraft version off intentional? 

Yeah, this is the docs for api 7. No need to worry about future versions of the api.

Grauldon (approved):

ST-DDT (approved):

PR #815, opened 2019-07-27
Updated dependency and lock file

felixoi:


package-lock.json
```diff
@@ -1507,7 +1639,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
Grauldon, line 1643: I am okay with `"optional": true` but I know Inscrutable had some concerns with it in the automatic PR. I point them out to ensure they are not an oversight.

Grauldon (approved):
Just the one verification.

package-lock.json
felixoi: Completely autogenerated by npm, so it's fine.

PR #820, opened 2019-10-14
Update guidelines.rst

Grinch:


ST-DDT (approved):

source/ore/guidelines.rst
```diff
@@ -185,8 +187,16 @@ player data, or map data) to external systems, the information collected must be
 Execution of Downloaded Code
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-This is a security risk we will not tolerate. This includes downloading jar or class files, generation of bytecode
-from downloaded sources, and execution shell scripts.
+We cannot ensure that content that is downloaded and executed at runtime is safe and complies with our guidelines. Any project that performs downloads and execution of code will have warnings on said project page
+and a warning prior to download to ensure users know the risk.
```
Inscrutable: If you can keep the line lengths < 120 it makes it easier, It's also one of the docs guidelines 📜 
I'd replace `said` with `the` to keep it simple. Also, do we mean "will" or "must" have warnings?

dualspiral: Will - it’s something we do so will happen 100% of the time, not something they do.

PR #840, opened 2020-06-02
Update notes on LagGoggle/TickCentral incompatibilties

LordRalex:
Due to constant issues with this set of mods, it's easier to just mark them as "avoid using when possible" and add relevant links. While there are times when you can have a working version, it's hit and miss. 

https://github.com/SpongePowered/SpongeForge/issues/3171 is a recent issue with their "new" solution to get around issues, and it is having issues.

### DIFF
--- a/source/server/spongineer/incompatible.rst
+++ b/source/server/spongineer/incompatible.rst
@@ -85,10 +85,11 @@ Just Enough IDs (JEID)
 - Solution: Update to version 1.0.3-54 or newer.
 
 LagGoggles/TickCentral
-~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~
 - Versions: All
-- Problem: Crash on startup (mixin conflict) or .
-- Solution: Due to the very invasive nature of these mods, it's hard for SpongeForge to run properly alongside them. Only use LagGoggles/TickCentral when you need to debug issues, and do not leave it on your server for production constant use.
+- Problem: Crash on startup (mixin conflict).
+- Solution: Due to the very invasive nature of these mods, SpongeForge cannot run properly alongside them. 
+  Only use these mods when you need to debug issues, and do not leave it on your server for production constant use.
 - Issues: https://github.com/SpongePowered/SpongeForge/issues/3171
 
 MystCraft
--- a/source/server/spongineer/incompatible.rst
+++ b/source/server/spongineer/incompatible.rst
@@ -88,9 +88,7 @@ LagGoggles/TickCentral
 ~~~~~~~~~~~~~~~~~~~~~~
 - Versions: All
 - Problem: Crash on startup (mixin conflict).
-- Solution: Due to the very invasive nature of these mods, SpongeForge cannot run properly alongside them. 
-  Only use these mods when you need to debug issues, and do not leave it on your server for production constant use.
-- Issues: https://github.com/SpongePowered/SpongeForge/issues/3171
+- Solution: No compatibility at present. Choose one or the other.
 
 MystCraft
 ~~~~~~~~~
--- a/source/server/spongineer/incompatible.rst
+++ b/source/server/spongineer/incompatible.rst
@@ -84,11 +84,12 @@ Just Enough IDs (JEID)
 - Problem: Crash on startup (mixin conflict).
 - Solution: Update to version 1.0.3-54 or newer.
 
-LagGoggles
+LagGoggles/TickCentral
 ~~~~~~~~~~
-- Versions: Older than 1.12.2-5.3-113
-- Problem: Crash on startup (mixin conflict).
-- Solution: Update to version 1.12.2-5.3-113 and install *TickCentral* core-mod.
+- Versions: All
+- Problem: Crash on startup (mixin conflict) or .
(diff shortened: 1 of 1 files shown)


source/server/spongineer/incompatible.rst
```diff
@@ -84,11 +84,12 @@ Just Enough IDs (JEID)
 - Problem: Crash on startup (mixin conflict).
 - Solution: Update to version 1.0.3-54 or newer.
 
-LagGoggles
+LagGoggles/TickCentral
 ~~~~~~~~~~
```
phit: need to match these to the titlelength

```
LagGoggles/TickCentral
~~~~~~~~~~~~~~~~~~~~~~
```

source/server/spongineer/incompatible.rst
```diff
@@ -84,11 +84,12 @@ Just Enough IDs (JEID)
...
-- Versions: Older than 1.12.2-5.3-113
-- Problem: Crash on startup (mixin conflict).
-- Solution: Update to version 1.12.2-5.3-113 and install *TickCentral* core-mod.
+- Versions: All
+- Problem: Crash on startup (mixin conflict) or .
```
ST-DDT: Incomplete sentence after `or`

source/server/spongineer/incompatible.rst
```diff
@@ -84,11 +84,12 @@ Just Enough IDs (JEID)
...
+- Solution: Due to the very invasive nature of these mods, it's hard for SpongeForge to run properly alongside them. Only use LagGoggles/TickCentral when you need to debug issues, and do not leave it on your server for production constant use.
```
ST-DDT: Line length limit 120 characters.

ST-DDT (changes requested):

source/server/spongineer/incompatible.rst
LordRalex: Is that a limit of the docs itself or just guidelines? Not sure we can easily explain this with 120 characters.

phit: you can just add linebreaks

```rst
- Solution: Due to the very invasive nature of these mods, it's hard for SpongeForge to run properly alongside them. 
  Only use LagGoggles/TickCentral when you need to debug issues, and do not leave it on your server for production 
  constant use.
```

Spongy:
A preview for this pull request is available at https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b22f0b5/.

- modified: https://cdn.rawgit.com/Spongy/SpongeDocs-PRs/b22f0b5/server/spongineer/incompatible.html

_Since the preview frequently changes, please link to [this comment](https://github.com/SpongePowered/SpongeDocs/pull/840#issuecomment-637723139), not to the direct url to the preview._

ST-DDT (approved):
LGTM

source/server/spongineer/incompatible.rst
```diff
@@ -84,11 +84,13 @@ Just Enough IDs (JEID)
...
 
-LagGoggles
-~~~~~~~~~~
-- Versions: Older than 1.12.2-5.3-113
+LagGoggles/TickCentral
+~~~~~~~~~~~~~~~~~~~~~~
+- Versions: All
 - Problem: Crash on startup (mixin conflict).
-- Solution: Update to version 1.12.2-5.3-113 and install *TickCentral* core-mod.
+- Solution: Due to the very invasive nature of these mods, SpongeForge cannot run properly alongside them.
```
Inscrutable: To be honest, we don't really have to dig into the nuts-and-bolts of why it goes wonky. I think this could use some diplomatic rephrasing, too.
"We suggest only using this mod when needed to debug issues. When used alongside SpongeForge, compatibility and game stability cannot be assured. See here for more info [Link]"

I'd like a more clear-cut solution, like "do not use with SpongeForge", but at least it gets around keeping a table of supposedly compatible versions.

source/server/spongineer/incompatible.rst
```diff
@@ -84,11 +84,13 @@ Just Enough IDs (JEID)
 - Problem: Crash on startup (mixin conflict).
 - Solution: Update to version 1.0.3-54 or newer.
 
-LagGoggles
-~~~~~~~~~~
-- Versions: Older than 1.12.2-5.3-113
+LagGoggles/TickCentral
```
Inscrutable, line 87: Isn't `Tickcentral` the coremod? We can just list that as a dependency, same as the others

Inscrutable (commented):
Grumble whinge moan etc. 
I didn't much like the phrasing, and seek clarity. 
Good job nonetheless. Let's kick this can down the road.

source/server/spongineer/incompatible.rst
LordRalex: TickCentral is, but older versions of LagGoggles before TickCentral came about also have problems with SpongeForge. I figured listing it explicitly is a better method to read than trying to list dependencies.

source/server/spongineer/incompatible.rst
LordRalex: I'll just make it a "do not use" like mystcraft, because there isn't a real way to run both at once since they crash.

source/server/spongineer/incompatible.rst
Inscrutable: The only reason I list them is so users know what else to remove. It'll do for now.

PR #846, opened 2020-09-04
Add ForgeEssentials to incompatibility list

ItsDoot:
First time docs PR; go easy on me!


Inscrutable (approved):
Excellent. We could note the Sponge version where the mixin change happened too, but it's a while back and we don't want to bog the doc down with SF versions.

ST-DDT (approved):

PR #805, opened 2019-05-25
Upgrading all the dependencies

phit:
Goal is to get a clean npm audit, but really updating all the things and replacing gulp-webserver with gulp-connect as the former hasn't seen a release in 4 years and [the maintainer said they no longer have the time to update it](https://github.com/schickling/gulp-webserver/issues/133#issuecomment-442847800).

todo:

~~rewrite gulpfile~~
(maybe see about updating sphinx, for now this is the node deps only)

PR #848, opened 2020-09-27
Update dependencies

felixoi:
This contains a huge update of sphinx.
I don't know why this wasn't done before, the only thing i couldn't test is sphinx-intl.
So tell me when this is broken.


phit:
> I don't know why this wasn't done before, the only thing i couldn't test is sphinx-intl.

at the time of https://github.com/SpongePowered/SpongeDocs/pull/805, jdlinker hadn't been updated for sphinx 2.x and broke

ST-DDT:
If you change hocon (`guess`) to `text`, then you should update it in the contributing guide as well:
https://github.com/SpongePowered/SpongeDocs/blame/stable/source/contributing/spongedocs.rst#L62

stephan-gh:
> I don't know why this wasn't done before, the only thing i couldn't test is sphinx-intl.

Well, there is a good chance that this will break something because the localization stuff tends to be rather wonky. But I suppose the only way to figure that out is to merge it and let Jenkins give it a try.

I would also make sure that sponge-docs-theme works/looks properly with the newer Sphinx version.

felixoi:
> > I don't know why this wasn't done before, the only thing i couldn't test is sphinx-intl.
> 
> Well, there is a good chance that this will break something because the localization stuff tends to be rather wonky. But I suppose the only way to figure that out is to merge it and let Jenkins give it a try.
> 
> I would also make sure that sponge-docs-theme works/looks properly with the newer Sphinx version.

I tested the sponge-docs-theme locally and it seems to work. Anyway it needs a new release on pypi. I'd btw like to move the maintainer from you to an official spongepowered account on pypi and automate the release using github actions (I think my discord message didn't reach you).

stephan-gh:
> I tested the sponge-docs-theme locally and it seems to work. Anyway it needs a new release on pypi. I'd btw like to move the maintainer from you to an official spongepowered account on pypi and automate the release using github actions (I think my discord message didn't reach you).

Yeah I don't check Discord very often and sadly no one ever made the IRC bridge work again. :(  
I somehow thought I did transfer those to a different account already... Weird.

PR #857, opened 2020-12-02
API 8 Command Docs

dualspiral:
Making a start. This will be fleshed out in chunks when I get inspiration to do it.

All the API 7 command docs are of no use to API 8 and should not be used as a guide as to how I'll approach the API 8 command docs. Comments are welcome on the substance while I write it as it'll only inform the style later on.

If things end up getting re-arranged on the docs front, just let me know and I'll rejig along with it.

(I think I also have the heading styles the wrong way around - I'll fix them later, Spinx seems to be broken on my comptuer at the moment so I'm flying blind!)


Inscrutable:
It seems the javadoc link to `CommandManager.Mutable` is not working.

dualspiral:
Oof, must have got the syntax wrong or something. I'll fix that when I sort out writing the last parts of this.

PR #860, opened 2021-01-16
Update Incompatible Mods List

Irgendwer01:
The Mod Laggoggles/Tickcentral is Compatible with API 7.2.0 or newer and the Mod Mystcraft is Compatible when changing B:useconfigs=false to true in the config/mystcraft/core.cfg file so this PR updates the list to tell people how to fix.


ST-DDT (approved):
Formatting wise this looks okay. I cannot verify the content though.

Irgendwer01:
@ST-DDT actually, the compatibility fix for Mystcraft i found here https://github.com/SpongePowered/SpongeForge/issues/2704#issuecomment-750485690

PR #754, opened 2018-08-26
Add tools for maintaining the docs

ST-DDT:
Currently everyone maintains a private copy of tools that he/she needs for maintenance tasks.

This is bad in multiple ways:

* Everyone has to write it himself
* Improvements don't get shared
* Team-members tend to change from time to time
* It cannot be used in automation

Thats why I started this attempt to publish the required sources in the repo as well.

------------------------------------------------------------

## Existing tools

* [x] Javadoc Checker - #640 #666 
* [x] Compile Code Checker - #695
* [x] ~`sponge-conf.rst` - Definition List Generator - #718 #732~ (different PR)

~@NariKiro It would be nice if you could upload your sources for the Definition List Generator somewhere/here (once the remaining questions have been resolved)~

------------------------------------------------------------

## Remaining questions

@SpongePowered/docs 

* [x] Store it in the Docs repo itself or in a separate repo/repo per tool?
  * In the repo
* [x] Which coding conventions should be used?
  * Sponge's but less strict
* [x] Restrictions on used applications such as jdk or build tools?
  * None
* [x] Which license should be used?
  * Any free license, preferable MIT or unlicense
* [x] ~Did I forget something?~

PR #875, opened 2021-07-25
Update jdlinker and try out new feature

12awsomeman34:
The jdlinker has just undergone a massive rewrite (some bugs may be expected), but also a new feature has been added. A shorthand method to importing javadoc links. Instead of typing the full ``org.spongepowered.api``, you can simply type ``sponge:`` and the jdlinker will automatically convert that to the full package for you. Let me know what you think.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 sphinx==3.2.1
 sphinx-intl==2.0.1
-sphinx-JDLinker==1.4
+sphinx-JDLinker==2.0.1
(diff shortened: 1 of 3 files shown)


ST-DDT (approved):
Looks good to me (and `works for me`).

But I'm not sure whether we should use the shorthand notation,
- as this makes it harder to verify whether the imports are still valid (#754)
- are generally harder to insert (I usually just use `copy qualified name` in my IDE)
- and I expect new/external contributors to strugle with it/use consistent

PR #888, opened 2022-01-12
update source/plugin/project/gradle to api-8

### LINKED ISSUE #877: Gradle page needs updating
The links to maven repos and gradle plugin info at /source/plugin/project/gradle.rst are wrong.
It has moved from org.spongepowered.plugin to org.spongepowered.gradle.plugin
[Thanks to Cableguy20 for letting us know).

This could get bundled with other PR's directed at the same page.

ImMorpheus:
The minimal example is based on https://github.com/SpongePowered/sponge-plugin-template/blob/api-8/build.gradle.kts

Fix https://github.com/SpongePowered/SpongeDocs/issues/877

### DIFF
--- a/source/plugin/project/gradle.rst
+++ b/source/plugin/project/gradle.rst
@@ -17,26 +17,30 @@ only need to update your plugin version in one file.
   version together with SpongeGradle_. :ref:`The Gradle section of the build systems page <gradle-setup>` explains how
   to setup Gradle on your computer.
 
+.. tip::
+  The `Sponge plugin template <https://github.com/SpongePowered/sponge-plugin-template/>`__ is a `Template repository <https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-repository-from-a-template>`__
+  that demonstrate SpongeGradle.
+
 Below is a simple template that should be usable for most plugins. **Make sure to replace the group with the group ID
 you have chosen before.**
 
-.. code-block:: groovy
+.. code-block:: kotlin
 
     plugins {
-        'java-library'
-        id("org.spongepowered.gradle.plugin") version "2.0.0"
+        r"""`java-library`
+        id("org.spongepowered.gradle.plugin") version "2.0.1"
     }
 
     // TODO: Change the following to match your information
-    group = 'com.example'
-    version = '1.0.0-SNAPSHOT'
+    group = "com.example"
+    version = "1.0.0-SNAPSHOT"
 
     repositories {
         mavenCentral()
     }
 
     sponge {
-        apiVersion("8.0.0-SNAPSHOT")
+        apiVersion("8.0.0")
         license("CHANGEME")
         loader {
             name(PluginLoaders.JAVA_PLAIN)
@@ -78,7 +82,7 @@ By default, the Gradle plugin will contribute the **plugin name**, **plugin vers
 to :doc:`/plugin/plugin-meta` with defaults defined in the project properties. It is also possible to override these if
 you want to specify them manually:
 
-.. code-block:: groovy
+.. code-block:: kotlin
 
     sponge {
         apiVersion("8.0.0-SNAPSHOT")
@@ -116,18 +120,17 @@ Without SpongeGradle
 Generally, everything necessary to compile a Sponge plugin using Gradle can be done by simply adding the SpongeAPI
 dependency to your project:
 
-.. code-block:: groovy
+.. code-block:: kotlin
 
     repositories {
         mavenCentral()
-        maven {
-            name = 'sponge-repo'
-            url = 'https://repo.spongepowered.org/repository/maven-public/'
+        maven("https://repo.spongepowered.org/repository/maven-public/") {
+            name = "sponge"
         }
     }
 
     dependencies {
-        compile 'org.spongepowered:spongeapi:8.0.0'
+        api("org.spongepowered:spongeapi:8.0.0")
     }
 
 .. _SpongeGradle: https://github.com/SpongePowered/SpongeGradle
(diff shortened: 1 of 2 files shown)


source/plugin/project/gradle.rst
```diff
@@ -23,7 +23,7 @@ you have chosen before.**
 .. code-block:: groovy
 
     plugins {
-        `java-library`
+        'java-library'
```
dualspiral: I thought it was meant to be backticks?

dualspiral: Oh, I see, you got a warning - can it just be escaped though? We must not provide incorrect code, which I think this would be.

zml2008: in Groovy buildscript, this would be:

```groovy
plugins {
  id 'java-library'
}
```

The example with backticks is only correct for Kotlin buildscripts:

```kotlin
plugins {
  `java-library`
}
```

If you wanted an example that would work in either, something like:

```gradle
plugins {
  id("java-library")
}
```

would work, but is not idiomatic for either language.

dualspiral: Right - while the code block is marked as "groovy", the block written here is kotlin anyway - especially as it seems to have been copied from the template as noted in the description.

That said, I think this page needs a broader update than just updating the code examples though - for example, it talks about autofilling the plugin name and version, not sure it's true for the former at least. We shouldn't mix kotlin and groovy either which seems to be happening here - choose one or the other (or both), though I would also consider pointing at the template repo.

source/plugin/project/gradle.rst
```diff
@@ -27,50 +23,54 @@ you have chosen before.**
...
+            name(PluginLoaders.JAVA_PLAIN)
+            version("1.0")
+        }
+        plugin("example") {
+            description("Just testing things...")
+            entrypoint("org.spongepowered.example.Example")
+        }
+    }
+
+    val javaTarget = 8 // Sponge targets a minimum of Java 8
```
dualspiral, line 55: `val` here is a Kotlin keyword, but the rest of this script is groovy, so you wouldn't have `val` here (you _might_ choose to use `def`, but I probably wouldn't here).

Make sure that the examples are consistent in the language they are written in.

source/plugin/project/gradle.rst
```diff
@@ -81,25 +81,27 @@ you want to specify them manually:
...
+            displayName("Example")
+            entrypoint("org.spongepowered.example.Example")
+            description("Just testing things...")
+            links {
+                homepage("https://spongepowered.org")
+                source("https://spongepowered.org/source")
+                issues("https://spongepowered.org/issues")
+            }
+            contributor("Spongie") {
+                description("Lead Developer")
```
Inscrutable, line 104: Clearly I need to amend the "Staff" page to reflect the power of Spongie 😁

Inscrutable (approved):
If you deal with dual's dilemmas I'd be delighted. There's no more I can usefully add.

source/plugin/project/gradle.rst
ImMorpheus: Changed to kotlin and added a tip about the template repository.

I've also setup a project using the template repo and made sure the version/description are being autofilled in the sponge_plugins.json

source/plugin/project/gradle.rst
ImMorpheus: Changed to kotlin

source/plugin/project/gradle.rst
dualspiral: Far better, thanks!

source/plugin/project/gradle.rst
```diff
@@ -21,85 +17,95 @@ only need to update your plugin version in one file.
...
 Below is a simple template that should be usable for most plugins. **Make sure to replace the group with the group ID
 you have chosen before.**
 
-.. code-block:: groovy
+.. code-block:: kotlin
 
     plugins {
-        id 'java'
-        id 'org.spongepowered.plugin' version '0.11.3'
+        r"""`java-library`
```
zml2008, line 30: `r"""`?

source/plugin/project/gradle.rst
```diff
@@ -114,18 +120,17 @@ Without SpongeGradle
 Generally, everything necessary to compile a Sponge plugin using Gradle can be done by simply adding the SpongeAPI
 dependency to your project:
```
zml2008, line 121: This is not as true anymore -- you'd have to manually write out a metadata file

source/plugin/project/gradle.rst
```diff
@@ -114,18 +120,17 @@ Without SpongeGradle
...
-            url = 'https://repo.spongepowered.org/repository/maven-public/'
+        mavenCentral()
+        maven("https://repo.spongepowered.org/repository/maven-public/") {
+            name = "sponge"
         }
     }
 
     dependencies {
-        compile 'org.spongepowered:spongeapi:8.0.0'
+        api("org.spongepowered:spongeapi:8.0.0")
```
zml2008:
```suggestion
        compileOnlyApi("org.spongepowered:spongeapi:8.0.0")
```

zml2008 (changes requested):

source/plugin/project/gradle.rst
Inscrutable: Or use https://github.com/SpongePowered/SpongeDocs/blob/stable/source/plugin/plugin-meta.rst#the-sponge_pluginsjson-format

Inscrutable:
Bump Bump. @ImMorpheus can you resolve the last few bits? I'd love to merge this.

PR #892, opened 2022-01-28
Document support for the Java platform logger

zml2008:
This is a companion to SpongePowered/Sponge#3608, targeting `api-9`.


ST-DDT (approved):
Looks good to me, but maybe we should create an API-9 branch (like we did for api8)

felixoi:
Don't merge this to stable. Create an api-9 branch please.

Inscrutable:
I think we can safely abandon the idea of API-9 docs, we barely had any releases for it.
This can be bumped forward to Sponge API-10.

ST-DDT:
Since we are way past api-9 should we merge it now or is it obsolete?

PR #898, opened 2022-03-16
Add Log4J error to mod imcompatibilities

dualspiral:
Not strictly the right place for it, but a place people look a lot.

### DIFF
--- a/source/server/spongineer/incompatible.rst
+++ b/source/server/spongineer/incompatible.rst
@@ -25,6 +25,23 @@ to rename the SpongeForge file, to make it first in the mod loading order. This 
 SpongeForge jar file name with ``aaa_``, giving a filename like ``aaa_spongeforge-1.12.2-2838-7.2.1-RC4011``. This makes it
 load the Sponge version of Mixin first, and can resolve many problems in one step.
 
+Log4J error
+~~~~~~~~~~~
+
+Though this is not strictly a mod error, it is often confused with being one. You may see the following line in your 
+crash log:
+
+.. code-block:: none
+
+    java.lang.NoSuchMethodError: org.apache.logging.log4j.spi.LoggerContextFactory.isClassLoaderDependent()Z
+
+If you do, **update to Forge 2860**. It is vitally important that you do so as Forge 2860 patches a severe vulnerability
+that was discovered in Log4J, a component that Minecraft uses, and can expose your clients as well as your server to 
+remote code execution vulnerabiltiies.
+
+If you update Forge by installing locally and copying Forge to a remote server, make sure you also copy the ``libraries`` 
+directory across, as this will contain a patched version of Log4J.
+
 FarSeek
 ~~~~~~~
 - Dependents: Streams mod


source/server/spongineer/incompatible.rst
```diff
@@ -25,6 +25,23 @@ to rename the SpongeForge file, to make it first in the mod loading order. This
...
+Though this is not strictly a mod error, it is often confused with being one. You may see the following line in your 
+crash log:
+
+.. code-block:: none
+
+    java.lang.NoSuchMethodError: org.apache.logging.log4j.spi.LoggerContextFactory.isClassLoaderDependent()Z
+
+If you do, **update to Forge 2860**. It is vitally important that you do so as Forge 2860 patches a severe vulnerability
+that was discovered in Log4J, a component that Minecraft uses, and can expose your clients as well as your server to 
+remote code execution vulnerabiltiies.
```
Inscrutable, line 40: I spy a wandering `i` in `vulnerabiltiies`.

Imconceivable,

Inscrutable (changes requested):
This needs to target API-7, not API-8. Pretty sure API-8 needs a *different* version of Forge (36.2.20+) - see https://gist.github.com/TheCurle/f15a6b63ceee3be58bff5e7a97c3a4e6

source/server/spongineer/incompatible.rst
```diff
@@ -25,6 +25,23 @@ to rename the SpongeForge file, to make it first in the mod loading order. This
...
+~~~~~~~~~~~
+
+Though this is not strictly a mod error, it is often confused with being one. You may see the following line in your 
+crash log:
+
+.. code-block:: none
+
+    java.lang.NoSuchMethodError: org.apache.logging.log4j.spi.LoggerContextFactory.isClassLoaderDependent()Z
+
+If you do, **update to Forge 2860**. It is vitally important that you do so as Forge 2860 patches a severe vulnerability
```
ST-DDT, line 38: Maybe add a `or later` to the forge version?

source/server/spongineer/incompatible.rst
```diff
@@ -25,6 +25,23 @@ to rename the SpongeForge file, to make it first in the mod loading order. This
...
+that was discovered in Log4J, a component that Minecraft uses, and can expose your clients as well as your server to 
+remote code execution vulnerabiltiies.
+
+If you update Forge by installing locally and copying Forge to a remote server, make sure you also copy the ``libraries`` 
+directory across, as this will contain a patched version of Log4J.
```
ST-DDT, line 43: Maybe instead of `copy` use `replace`? To ensure the broken version is really gone and not loaded by anything.

ST-DDT (changes requested):

dualspiral:
> This needs to target API-7, not API-8. Pretty sure API-8 needs a different version of Forge (36.2.20+)

Forgive me for the rambling, but there is are important reasons why I think it should still target the stable branch, as well as the API 7 branch (if there is one).

The ultimate problem is that we're still linking to this page, even though it's now "API 8". When I went and did a first pass tagging everything, I opted to change the page to say "SpongeAPI 7" specifically because if we blanked this page at this time, it would cause confusion. Additionally, we're still linking to the "stable" version of the page in other places, and I'm sure others are too.

Server owners who are trying to set up a server and look at the docs are also greeted with a large page with flags for languages. Nowhere is there an indication that these docs are for API-8. Because of how large it is, for most, it's not obvious there are old versions, and many new server owners won't necessarily link 7.4.0 to Minecraft 1.12.2. Let's face it, in general, server owners are just going to want to read some docs - and if they go to the incompatibilities page and can't see anything relavent, they're not necessarily going to know that they're looking at the wrong API version.

I think there is probably a larger conversation, outside of this ticket, about all of this. However, the above is why I updated that page with a header when I went through the pages, rather than wiping it.

Inscrutable:
I have started this discussion in the Spongedocs discord channel so we can get some more input.
My preference is to put a nice banner near the top, linking the Minecraft 1.12.2 (API-7) page in nice friendly words. API-8 has its own incompatibilities already and we need somewhere to put them.

PR #911, opened 2022-06-05
Farewell kashike

Inscrutable:
- Retired kashike from all roles
- Retired Ore Review team
- Retired Minecrell, ewoutvs_

There are several moderators that probably need to be moved to retired too (gravityfox and Lemonous).
I'm not sure if Nari is still involved with SpongePowered either.


ST-DDT (approved):

source/about/staff.rst
```diff
@@ -242,8 +198,14 @@ here to thank them and acknowledge their valuable contributions.
 +-----------------------------------+------------------------------------+
 | Disconsented                      | Moderator                          |
 +-----------------------------------+------------------------------------+
+| doot                              | Ore Reviewer                       |
```
dualspiral: doot, mosemister, RedNesto, WillBAnders and Yeregorix aren't retired - they're still moderators for us.

ST-DDT (approved):

PR #956, opened 2023-04-18
Updated items

mosemister:
Updates the item docs to api 8

## Main Changes

To be fair, items havent changed too much in terms of text describing the api from 7 to 8, so there isn't much to talk about

### Item properties
While these have been removed in API 8 with it being changed to ``Keys``, I am wanting to change this to item type ``keys``. Just need to find a good example.

## On Hold

~~This is being blocked by https://github.com/SpongePowered/Sponge/issues/3845 (unless there is no keys to begin with for ItemType)~~ Fixed

Im wanting to double check the changes have applied. Need some free time

### DIFF
--- a/source/plugin/items/creating.rst
+++ b/source/plugin/items/creating.rst
@@ -2,14 +2,9 @@
 Creating an ItemStack
 =====================
 
-.. warning::
-    These docs were written for SpongeAPI 7 and are likely out of date. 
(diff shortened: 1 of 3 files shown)


Inscrutable:
We should clarify the keys issue at the next SPAM. This needs to move forward.

gabizou (approved):

PR #966, opened 2023-05-22
Updates Setup Pages

mosemister:
It seems the setup pages have already had a good go at before with only the Gradle setup page being out of date. 

This PR updates the Gradle page to the latest SpongeGradle setup

### DIFF
--- a/source/plugin/project/gradle.rst
+++ b/source/plugin/project/gradle.rst
@@ -4,10 +4,6 @@ Setting Up Gradle
 
 .. _using-spongegradle:
 
-.. warning::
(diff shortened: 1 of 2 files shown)


mosemister:
Due to Gradle preferring kotlin as its gradle.build file, the gradle build files have been rewritten in it. Luckly it seems there isnt any real change. It also appears ``kotlin`` codeblock isnt supported so the codeblocks are still in groovy and highlights are still working in its case

ST-DDT (commented):
This looks alright to me, but I currently dont have time for an in depth check.

ST-DDT (approved):

PR #978, opened 2023-07-21
Events for API 8

mosemister:
Adds the event docs for API 8.

Generally events havent changed that much when compared to API 7. 

It seems someone had gone though before me and updated some of the imports and method names, so this one isnt that big


ST-DDT (approved):
I haven't checked this in depth, but looks good to me.

Inscrutable (approved):

PR #982, opened 2023-08-31
(API 8) Minor updates

mosemister:
This branch was meant for the economy, however this seems to be mostly updated anyway. So this branch became minor edits to the docs to provide more information. 

The following was added:

- A note to implement the ``EconomyTransactionEvent`` (Some Economy plugins don't fire this event -> could be a assumption that Sponge was going to do it for them which would be incorrect)
- Registering your Economy Service (In API8 its now done by events)
- Fixed player only command example
- Minor edit to specify that in a command register example, the two commands are different types and that the first command is referencing an example code block from the same page
- Clarified how to properly run a command via a player
- Clarified in a code block how to get the command manager

### DIFF
--- a/source/plugin/commands/commandmanager.rst
+++ b/source/plugin/commands/commandmanager.rst
@@ -63,4 +63,4 @@ Sending a Command As a Player With Console Permissions
 
     Sponge.server().causeStackManager().pushCause(player);
     CommandManager cmdManager = Sponge.server().commandManager();
-    cmdManager.process(console, player, "kill Nitch");
+    cmdManager.process(console, player, "kill Notch");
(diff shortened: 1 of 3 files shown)


source/plugin/commands/commandmanager.rst
```diff
@@ -11,20 +11,56 @@ The :javadoc:`CommandManager` stands as the manager for watching what commands g
...
 .. code-block:: java
 
-    cmdManager.process(Sponge.systemSubject(), player, "kill Nitch");
\ No newline at end of file
+    ServerPlayer player;
+    SystemSubject console = Sponge.systemSubject();
+
+    Sponge.server().causeStackManager().pushCause(player);
+    CommandManager cmdManager = Sponge.server().commandManager();
+    cmdManager.process(console, player, "kill Nitch");
```
ST-DDT: Is this a typo or intentional?

mosemister: That was there before. Ill correct it

mosemister: Fixed

ST-DDT (approved):

PR #992, opened 2023-12-31
(API 8+) Updated API page to state API 8 EOL

mosemister:
Updates the sponge releases to state API 8 is out and API 11 is maybe targetting 1.20.4.

The new dates of the sponge 8.x releases were originally based on when the actual release occured to the last github commit. 

it is now based on the first github commit to when that version released


ST-DDT:
> The new dates of the sponge 8.x releases were originally based on when the actual release occured to the last github commit.
> 
> it is now based on the first github commit to when that version released

Was that first commit version for v8.0.0 usable at all?

mosemister:
> > The new dates of the sponge 8.x releases were originally based on when the actual release occured to the last github commit.
> > it is now based on the first github commit to when that version released
> 
> Was that first commit version for v8.0.0 usable at all?

Yep. Thats the exception to the rule i did, whereby that link was taken from the first download link for api 8

ST-DDT:
What speaks against the continued use of the first release?

mosemister:
> What speaks against the continued use of the first release?

Mind rephrasing?

ST-DDT:
Seems that the idiom didn't translate well.

AFAICT the previous date represent the actual release date of that version (v8.0.0 not snapshot) .
The column name also is `Release Date`.
Why switching from the actual release to a unstable dev build publication?

If we change the semantics of the dates, we should also change the column name.

mosemister:
> Seems that the idiom didn't translate well.
> 
> AFAICT the previous date represent the actual release date of that version. The column name also is `Release Date`. Why switching from the actual release to a unstable dev build publication?
> 
> If we change the semantics of the dates, we should also change the column name.

Ah fair. The reason why is unlike every other date, when it came to 8.2 it makes it look like we released 8.2 without any support. I can change it back. But didnt seem right

ST-DDT:
Starting a new major version doesn't automatically end support for the previous version. 
IIRC there are decisions during the team meetings to end support for a version.
For minor versions this doesn't apply due to the additional work involved.

I don't know what is was it like for 8.2.0. But maybe the actual support for API v8 ended with that last release.
(End of support does not imply end of usability)

mosemister:
> Starting a new major version doesn't automatically end support for the previous version. IIRC there are decisions during the team meetings to end support for a version. For minor versions this doesn't apply due to the additional work involved.
> 
> I don't know what is was it like for 8.2.0. But maybe the actual support for API v8 ended with that last release. (End of support does not imply end of usability)

Oh i know it doesnt mean end of usability. However i know a lot of people do believe that end of support means end of use ability and with it saying that on the date it was released was the same date it ended support. It doesnt inspire confidence in support

As for when api 8.2 ended support. It was decided during the snapshots for 8.2 that it would be the last before moving fully to api 10/11

ST-DDT:
Should we call it end of updates instead of end of support then?

mosemister:
> Should we call it end of updates instead of end of support then?

Thats a good shout. Ill change that

mosemister:
> Should we call it end of updates instead of end of support then?

Made the changes

ST-DDT (approved):

PR #994, opened 2024-01-03
Fix minor typos in Child Commands

BrendonButler:
Fixes subcommand example typo where `/main` is referenced instead of `/mail`.

Includes other minor typo fixes that were identified when making this change.

### DIFF
--- a/source/plugin/commands/childcommands.rst
+++ b/source/plugin/commands/childcommands.rst
@@ -8,13 +8,13 @@ Child Commands
     org.spongepowered.api.command.exception.ArgumentParseException
     java.lang.String
 
-The :javadoc:`Builder` supports hierachical command structure like this:
+The :javadoc:`Builder` supports hierarchical command structure like this:
 
 - ``/mail`` (parent command)
-    - ``/main send`` (child command)
-    - ``/main read`` (child command)
+    - ``/mail send`` (child command)
+    - ``/mail read`` (child command)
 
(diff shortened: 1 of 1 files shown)


source/plugin/commands/childcommands.rst
```diff
@@ -55,7 +55,7 @@ If a command has child commands, a :javadoc:`CommandExecutor`, set through
 :javadoc:`Builder#executor(CommandExecutor)` and the parameters for the executor are set to optional.
```
BrendonButler: The only potential grammar issue I could find is with this wording: 

> If a command has child commands, a `CommandExecutor`, set through `Builder#executor(CommandExecutor)` and the parameters for the executor are set to optional.

I'm unsure what is meant by this?

mosemister: Hi. This means that under normal circumstances, the ``executor`` method is a required method to be called when building a command. However when the command has children, it is not required. 

Essentially its a requirement to either set a executor or to set children of the command.

Hope that helps. 

Thanks for helping out the docs :)

BrendonButler: @mosemister how do the new changes look? If the before is preferred, this change can be reverted. Everything else regarding spelling and grammar/flow looks good to me--I can't identify any other issues with this page.

mosemister: Looks good. Approved

PR #1004, opened 2024-06-27
Fix typo descritption -> description in rawcommand.rst

JorelAli:
Fixes a typo in the Raw Command page where a variable was named `descritption` instead of `description`.


mosemister (approved):
Approved

ST-DDT (approved):

ST-DDT:
I think our CI is broken.

mosemister:
> I think our CI is broken.

It is. Has been for a while

felixoi:
> It is. Has been for a while

It's versioning.. We would either need to upgrade sphinx or pin some dependencies (sphinx requires them but doesnt pin them, so pip just installs lastest..). Currently no time available for me to do that

PR #1015, opened 2024-10-12
Document thread-safe connection event usage

aromaa:


source/plugin/event/connection.rst
```diff
@@ -0,0 +1,167 @@
...
+- Event ``ServerSideConnectionEvent.Auth`` is fired.
+    - Plugin A starts to load user data.
+    - Connection drops. Auth event is still being processed, disconnect event is queued.
+    - Plugin A finishes loading user data.
+- Event ``ServerSideConnectionEvent.Disconnect`` is fired.
+
+Best Practices
+==============
+Due to the nature of connection events being asynchronously, plugin developers **must** use thread-safe
+collections when storing, accessing and mutating related data.
```
mosemister, line 62: Due to many people not knowing there is thread safe collections, could we have it link to here (the official docs on concurrent collections). I know you have a example below, just a link to give more examples)

https://docs.oracle.com/javase/tutorial/essential/concurrency/collections.html

mosemister (changes requested):
Only done a quick scan with one comment so far. More may come when i have the time to properly load up your branch

